PR #46, opened 2020-08-25
Add configuration validation beforehand

### LINKED ISSUE #44: Don't panic when someone has passed an unsupported issue tracker
When someone specifies an unsupported issue tracker, a non-user friendly panic occurs. See #43 

Instead, validate the issue tracker & configuration beforehand & print a more user-friendly error message.

mehdy:
Fixes #44

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -26,6 +26,14 @@ const (
 	IssueTrackerRedmine              = "REDMINE"
 )
 
+var validIssueTrackers = []IssueTracker{
+	IssueTrackerJira,
+	IssueTrackerGithub,
+	IssueTrackerGitlab,
+	IssueTrackerPivotal,
+	IssueTrackerRedmine,
+}
+
 var windowsAbsolutePathPattern = regexp.MustCompile("^[A-Z]{1}:")
 
 // Local todocheck configuration struct definition
@@ -64,6 +72,27 @@ func NewLocal(cfgPath, basepath string) (*Local, error) {
 	removeCurrentDirReference(cfg.IgnoredPaths)
 
 	return cfg, nil
+}
+
+// Validate validates the values of given configuration
+func (l *Local) Validate() []error {
+	var errors []error
+
+	if err := l.validateIssueTracker(); err != nil {
+		errors = append(errors, err)
+	}
+
+	return errors
+}
+
+func (l *Local) validateIssueTracker() error {
+	for _, issueTracker := range validIssueTrackers {
+		if l.IssueTracker == issueTracker {
+			return nil
+		}
+	}
+	return fmt.Errorf("invalid issue tracker: %q is not supported. the valid issue trackers are: %v",
+		l.IssueTracker, validIssueTrackers)
 }
 
 func exists(filepath string) bool {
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -66,6 +66,34 @@ func NewLocal(cfgPath, basepath string) (*Local, error) {
...
+}
+
+func (l *Local) validateIssueTracker() error {
+	validIssueTrackers := []IssueTracker{
+		IssueTrackerJira,
+		IssueTrackerGithub,
+		IssueTrackerGitlab,
+		IssueTrackerPivotal,
+		IssueTrackerRedmine,
+	}
```
preslavmihaylov, line 87: Do you think it would be better to put this block somewhere close to the enum declaration?

Unfortunately, there is no way to foreach an enum in go, without this supplementary array, but we could at least make our lives easier by making it so when someone adds an enum in the future, the additional changes that have to be made is closer & easier to spot.

config/config.go
```diff
@@ -66,6 +66,34 @@ func NewLocal(cfgPath, basepath string) (*Local, error) {
 	return cfg, nil
 }
 
+// Validate validates the values of given configuration
+func (l *Local) Validate() []error {
+	var errors []error
+
+	if err := l.validateIssueTracker(); err != nil {
+		errors = append(errors, err)
+	}
```
preslavmihaylov, line 75: I like this approach with the `[]errors`. The only thing I would change in the future is invoking the validation funcs like this:
```
func (l *Local) validateFuncs(fs validateFunc) []errors { ... }

Then, the usage would look like:
func (l *Local) Validate() []errors {
    return validateFuncs(validateIssueTracker, validateFoo, validateBar)
}
```

No need for changes now as it is only a single validation func, but wanted to hear your thoughts.

main.go
```diff
@@ -31,6 +31,14 @@ func main() {
 		log.Fatalf("couldn't open configuration file: %s\n", err)
 	}
 
+	if errors := localCfg.Validate(); len(errors) > 0 {
+		for _, err := range errors {
+			log.Println(err)
+		}
+
+		os.Exit(1)
+	}
```
preslavmihaylov, line 40: Could you also add a new test scenario for this new feature?

A similar test can be seen [here](https://github.com/preslavmihaylov/todocheck/blob/master/testing/todocheck_test.go#L322)

To run the tests locally:
```
go test -v -count=1
```

preslavmihaylov (changes requested):
Thanks a lot for this PR!

I have a couple of comments & looking forward to merge after they're addressed.

config/config.go
mehdy: Totally agreed! I'll change it.

config/config.go
mehdy: Sure! I think it's a great way to deal with them. maybe we could do it if we had 3 or more validation functions to run! :+1:

main.go
mehdy: Thanks! I'll take care of it.

mehdy: Done.

config/config.go
mehdy: Done.

preslavmihaylov:
Thanks a lot for the PR, approving all changes & merging

preslavmihaylov (approved):

PR #49, opened 2020-08-27
Add offline url validation when auth type is offline

### LINKED ISSUE #36: Return error when no offline_url is specified with auth type offline
When the auth type is offline, there should also be an `offline_url` field in the configuration specified.

If such is not specified, no error is shown, instead, the application works but shows:
```
Please go to  and paste the offline token below:
Token:
```

mehdy:
Fixes #36

### DIFF
--- a/testing/test_configs/invalid_offline_url.yaml
+++ b/testing/test_configs/invalid_offline_url.yaml
@@ -1,0 +1,5 @@
+origin: http://127.0.0.1:51596
+issue_tracker: JIRA
+auth: 
+  type: offline
+  offline_url: INVALID_URL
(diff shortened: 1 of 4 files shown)


testing/test_configs/invalid_offline_url.yaml
```diff
@@ -0,0 +1,5 @@
+origin: http://127.0.0.1:51596
+issue_tracker: JIRA
+auth: 
+  type: offline
+  offline_url: INVALID_URL
```
preslavmihaylov, line 5: Nice! 

Could you also add a test for when auth type is offline & `offline_url` is completely missing?

mehdy: sure!

mehdy: This is done.

preslavmihaylov (approved):

Issue #59, opened 2020-09-09
Strip the trailing ".git" when auto-detecting config from an SSH-cloned git repo


preslavmihaylov:
When a repo is cloned using ssh, it has the `.git` suffix in its remote url in `.git/config`.

This is correctly detected from todocheck as a github configuration. However, all subsequent calls for issues fail as the github API doesn't consider `github.com/user/repo.git` the same as `github.com/user/repo`.

See example:
![image](https://user-images.githubusercontent.com/7642330/92556137-d5c67200-f271-11ea-80c8-655c2008b21f.png)

In the image above, I have a private repo setup, called `todocheck-mock` with issues 1 (closed) and 2 (open).
`todocheck` incorrectly detects both issues as non-existent as the API calls are not targeted towards the right repo.

preslavmihaylov:
Follow-up from #50 and #52 

@mehdy cc

preslavmihaylov:
Double-checked this again by instrumenting the app \w a logline showing the interpreted origin:
![image](https://user-images.githubusercontent.com/7642330/92568817-660fb180-f288-11ea-8de0-940c620d0b6f.png)

mehdy:
This is odd. :raised_eyebrow: 
I'll look into it.

PR #61, opened 2020-09-11
Fix issue tracker auto detector's regex

mehdy:
The new regex now detects `origin` and `host` better and totally ignores the existence of the `.git` part.

Fixes: #59


preslavmihaylov:
Thanks, I've tested your branch locally & it works as expected now!

I've also discovered another issue (#62) in the process.

preslavmihaylov (approved):
I've tested the branch locally & it now works as expected.

PR #65, opened 2020-09-23
Add contributing guide

### LINKED ISSUE #24: Write CONTRIBUTING guide
The guide should include instructions about local development - initial setup, building, running tests.

All steps are the standard go commands. The only caveat with running tests is that it requires a little bit more options:
```
go test -v -count=1 ./testing
```

This step also depends on running `go build` beforehand.

preslavmihaylov:
closes #24


mehdy (approved):
Thanks! This is really great.

PR #69, opened 2020-10-06
Fix links in CONTRIBUTING.md

codykaup:
I was looking through the repo and found some funky links in the contributing guide. Here's the quick fix to get those looking right!


mehdy (approved):
Thanks!

Issue #63, opened 2020-09-13
Show a warning if no apitoken is specified for any github repository


preslavmihaylov:
Github has a [rate limit in-place](https://developer.github.com/v3/#rate-limiting) for unauthenticated requests to their API which is 60 requests per hour.

For any production usage, this is a very low quota.
However, if the requests are authenticated (even if accessing public repositories), the quota is 5000 requests per hour, which should suffice for most use cases.

Therefore, a warning should be shown if the configuration lacks an `apitoken` for github repositories, regardless of whether the repository is private or not.

In scope of this issue is to include this warning in the README for public github repository setup as well.

You can reuse the mechanics of #62 for showing this warning.

codykaup:
I'd like to snag this one as well if it's still available!

preslavmihaylov:
@codykaup go ahead!

PR #70, opened 2020-10-07
Check access to GitHub repo on issuetracker initialization

### LINKED ISSUE #62: Return an error if github repository is not found or private
Currently, if someone specifies a valid private github repository without an `apitoken`, `todocheck` won't report any error but will instead report all linked issues in TODOs as not found.

The reason is that Github returns status code `404 Not Found` instead of `403 Unauthorized` when querying private repositories for privacy/security reasons.

In this situation, it takes quite a while to find out what the issue is until you realize you haven't specified the `apitoken` in your configuration.

It would be helpful for `todocheck` to do the following request upon initializing the configuration:
```
curl https://api.github.com/repos/{user}/{repo}
```

If this request returns `200`, the repository exists. If it returns `404`, it might not exist OR is private.

Create a **on-initialize** hook for issuetrackers, which can do arbitrary checks. For github, this is the check to do.
If it doesn't pass AND there is no auth token specified, print the error message & exit:
```
Github repository %s not found. Are you accessing a private repository without a token?
```

codykaup:
This PR closes #62 by making a quick request to the Github repo to ensure the user has access. If a 404 is returned, we log the output and exit early!

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,10 +60,6 @@ func main() {
 		log.Fatalf("couldn't get base url for origin %s & issue tracker %s: %s\n",
 			localCfg.Origin, localCfg.IssueTracker, err)
 	}
-	err = validation.Validate(localCfg.IssueTracker, localCfg.Origin)
-	if err != nil {
-		log.Fatal(err)
-	}
 
 	todoErrs := []*todocheckerrors.TODO{}
 	f := fetcher.NewFetcher(baseURL, localCfg.IssueTracker, authmiddleware.For(localCfg))
--- a/main.go
+++ b/main.go
@@ -66,9 +66,7 @@ func main() {
 	if runningTests != "true" {
 		err = validation.Validate(localCfg.IssueTracker, localCfg.Origin)
 		if err != nil {
-			if !errors.Is(err, issuetracker.ErrUnsupportedHealthCheck) {
-				log.Fatalf("Github repository %s not found. Are you accessing a private repository without a token?", localCfg.Origin)
-			}
+			log.Fatal(err)
 		}
 	}
 
--- a/main.go
+++ b/main.go
@@ -21,9 +21,6 @@ import (
 // set dynamically on build time. See Makefile for more info
 var version string
 
-// set dynamically on build time. If this value is "true", it will skip unnecessary external calls
-var runningTests string
-
 // TODO:
 // * Add a --closes option which indicates that an issue is to be closed as a result of a PR
 // * Add caching for task statuses
@@ -63,11 +60,9 @@ func main() {
 		log.Fatalf("couldn't get base url for origin %s & issue tracker %s: %s\n",
 			localCfg.Origin, localCfg.IssueTracker, err)
 	}
-	if runningTests != "true" {
-		err = validation.Validate(localCfg.IssueTracker, localCfg.Origin)
-		if err != nil {
-			log.Fatal(err)
-		}
+	err = validation.Validate(localCfg.IssueTracker, localCfg.Origin)
+	if err != nil {
+		log.Fatal(err)
 	}
 
 	todoErrs := []*todocheckerrors.TODO{}
--- a/main.go
+++ b/main.go
@@ -21,6 +21,9 @@ import (
 // set dynamically on build time. See Makefile for more info
 var version string
 
+// set dynamically on build time. If this value is "true", it will skip unnecessary external calls
+var runningTests string
+
 // TODO:
 // * Add a --closes option which indicates that an issue is to be closed as a result of a PR
 // * Add caching for task statuses
@@ -59,6 +62,14 @@ func main() {
 	if err != nil {
 		log.Fatalf("couldn't get base url for origin %s & issue tracker %s: %s\n",
 			localCfg.Origin, localCfg.IssueTracker, err)
+	}
+	if runningTests != "true" {
+		err = validation.Validate(localCfg.IssueTracker, localCfg.Origin)
+		if err != nil {
+			if !errors.Is(err, issuetracker.ErrUnsupportedHealthCheck) {
+				log.Fatalf("Github repository %s not found. Are you accessing a private repository without a token?", localCfg.Origin)
+			}
+		}
 	}
 
 	todoErrs := []*todocheckerrors.TODO{}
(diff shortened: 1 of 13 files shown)


codykaup:
@preslavmihaylov are these tests working for you locally? I didn't touch that part of the code so I wonder if it's something on my end or if the tests themselves are funky. 🤔

main.go
```diff
@@ -20,6 +21,9 @@ import (
 // set dynamically on build time. See Makefile for more info
 var version string
 
+// set dynamically on build time. If this value is "true", it will skip unnecessary external calls
+var runningTests string
```
codykaup, line 25: It was a bit tricky to run the full test suite without having a flag to skip the Github check. It doesn't feel the best but I'm open to suggestions if there's a better way!

preslavmihaylov (changes requested):
@codykaup, thanks a lot for putting together this PR.

Unfortunately, it turns out to be more convoluted than what it seemed in hindsight. See [my comment](https://github.com/preslavmihaylov/todocheck/issues/62#issuecomment-706503876) on the linked issue.

That's what is necessary to complete this while keeping the packages decoupled.

For example, in this PR you're making an http request in the `config` package, while the `fetcher` is the package which should wrap any kind of http calls.

codykaup:
> @codykaup, thanks a lot for putting together this PR.
> 
> Unfortunately, it turns out to be more convoluted than what it seemed in hindsight. See [my comment](https://github.com/preslavmihaylov/todocheck/issues/62#issuecomment-706503876) on the linked issue.
> 
> That's what is necessary to complete this while keeping the packages decoupled.
> 
> For example, in this PR you're making an http request in the `config` package, while the `fetcher` is the package which should wrap any kind of http calls.
> 
> If you still want to finish the PR after taking those changes into consideration, I am willing to assist in real-time if you'd want.
> 
> Otherwise, I'd suggest picking #63 which is a lot easier & doesn't require major refactoring.

Thanks for that comment, it gave me some good info to work from! I just pushed up the new layout to see what you think. Am I on the right track with what you were thinking?

preslavmihaylov (changes requested):
This is definitely in the right direction. There are some issues still left to address but merging this PR is definitely in hindsight.

I will be available to provide more active feedback these days to help getting it merged.

preslavmihaylov:
If the problem is that some testing configs include non-existent github repos, why don't you try changing them to a valid public github repo instead?

codykaup:
> If the problem is that some testing configs include non-existent github repos, why don't you try changing them to a valid public github repo instead?

Turns out, it was a lot easier than I thought. There were just 2 places I needed to update so this one is ready!

validation/validation.go
```diff
@@ -0,0 +1,26 @@
...
+	url, err := issuetracker.HealthcheckURL(issueTracker, origin)
+	if err != nil {
+		if errors.Is(err, issuetracker.ErrUnsupportedHealthCheck) {
+			return nil
+		}
+		return err
+	}
+
+	if !fetcher.IsHealthy(issueTracker, url) {
+		return errors.New("user does not have access to repo")
```
preslavmihaylov, line 22: Please include the repo URL which is problematic in the error message.

Additionally, this can be changed to:
```
permission denied for repository %s
```

preslavmihaylov (changes requested):

validation/validation.go
codykaup: No idea how I missed that. Updated!

testing/todocheck_test.go
```diff
@@ -331,6 +332,24 @@ func TestInvalidIssueTracker(t *testing.T) {
 	}
 }
 
+func TestValidGithubAccess(t *testing.T) {
+	err := validation.Validate(config.IssueTrackerGithub, "https://github.com/preslavmihaylov/todocheck")
+	if err != nil {
+		t.Errorf("%s", err)
+	}
```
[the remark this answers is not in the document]
codykaup: Good call. Just pushed that update!

preslavmihaylov:
The PR I mentioned previously - #73 has been merged to master. Feel free to pull & update yours.

codykaup:
@preslavmihaylov just a heads up, this is ready for another review! Thanks for all your help on this one!

validation/validation.go
```diff
@@ -1,27 +1,41 @@
...
+
+	url, err := issuetracker.HealthcheckURL(cfg.IssueTracker, cfg.Origin)
+	if err != nil {
+		if !errors.Is(err, issuetracker.ErrUnsupportedHealthCheck) {
+			errs = append(errs, err)
+		}
+	} else {
+		if !fetcher.IsHealthy(cfg.IssueTracker, url) {
+			errs = append(errs, fmt.Errorf("repository %s not found. Is the repository private? More info: https://github.com/preslavmihaylov/todocheck#github", url))
+		}
```
preslavmihaylov, line 38: Final touch.

For consistency with the rest of this function, could you encapsulate your new validation logic into a `validateIssueTrackerExists(cfg)`

validation/validation.go
```diff
@@ -1,27 +1,41 @@
...
+		errs = append(errs, err)
+	}
+
+	url, err := issuetracker.HealthcheckURL(cfg.IssueTracker, cfg.Origin)
+	if err != nil {
+		if !errors.Is(err, issuetracker.ErrUnsupportedHealthCheck) {
+			errs = append(errs, err)
+		}
+	} else {
+		if !fetcher.IsHealthy(cfg.IssueTracker, url) {
```
preslavmihaylov, line 36: This can be simplified from:
```
} else {
    if !fetcher.IsHealthy(cfg.IssueTracker, url) {
```

to
```
} else if !fetcher.IsHealthy(cfg.IssueTracker, url) {
```

preslavmihaylov (changes requested):

preslavmihaylov:
@codykaup this is the final touch & it's ready to go :)

validation/validation.go
codykaup: I ended up simplifying this flow a bit after extracting to an unexported function. I think it reads better but let me know what you think!

preslavmihaylov (approved):

preslavmihaylov:
Thanks @codykaup a lot for this contribution! It's officially in master 🎉

PR #72, opened 2020-10-15
Show warning if no token is provided on a Github repo

### LINKED ISSUE #63: Show a warning if no apitoken is specified for any github repository

codykaup:
This PR closes #63 by printing an error (in red) if the user is accessing a Github repo without an API token! I tried to make things pretty explicit so it's easier to reason about but let me know what you think!

### DIFF
--- a/authmanager/authmanager.go
+++ b/authmanager/authmanager.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/fatih/color"
 	"github.com/preslavmihaylov/todocheck/authmanager/authstore"
 	"github.com/preslavmihaylov/todocheck/common"
 	"github.com/preslavmihaylov/todocheck/config"
@@ -20,31 +19,21 @@ const (
 	pivotalAPITokenMsg = "Please go to https://www.pivotaltracker.com/profile, create a new API token & paste it here:\nToken: "
 	redmineAPITokenMsg = "Please go to %s/my/account, create a new API token & paste it here:\nToken: "
 
-	githubAPITokenWarning = "WARNING: Github has API rate limits for all requests which do not contain a token.\nPlease create a read-only access token to increase that limit.\nGo to https://developer.github.com/v3/#rate-limiting for more information."
-
 	authTokenEnvVariable = "TODOCHECK_AUTH_TOKEN"
 )
 
 // AcquireToken stores the issue tracker's auth token based on the auth type specified
 func AcquireToken(cfg *config.Local) error {
-	err := func() error {
-		switch cfg.Auth.Type {
-		case config.AuthTypeNone:
-			return nil
-		case config.AuthTypeAPIToken:
-			return acquireAPIToken(cfg)
-		case config.AuthTypeOffline:
-			return acquireOfflineToken(cfg.Auth)
-		default:
-			panic("unknown auth token type")
-		}
-	}()
-
-	if cfg.Auth.Token == "" {
-		showAuthWarning(cfg)
+	switch cfg.Auth.Type {
+	case config.AuthTypeNone:
+		return nil
+	case config.AuthTypeAPIToken:
+		return acquireAPIToken(cfg)
+	case config.AuthTypeOffline:
+		return acquireOfflineToken(cfg.Auth)
+	default:
+		panic("unknown auth token type")
 	}
-
-	return err
 }
 
 func acquireAPIToken(cfg *config.Local) error {
@@ -120,13 +109,4 @@ func extractBaseURL(origin string) string {
 	}
 
 	return fmt.Sprintf("%s//%s", tokens[0], tokens[1])
-}
-
-func showAuthWarning(cfg *config.Local) {
-	switch cfg.IssueTracker {
-	case config.IssueTrackerGithub:
-		fmt.Println(color.YellowString(githubAPITokenWarning))
-	default:
-		// noop
-	}
 }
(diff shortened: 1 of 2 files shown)


authmanager/authmanager.go
```diff
@@ -19,21 +20,31 @@ const (
 	pivotalAPITokenMsg = "Please go to https://www.pivotaltracker.com/profile, create a new API token & paste it here:\nToken: "
 	redmineAPITokenMsg = "Please go to %s/my/account, create a new API token & paste it here:\nToken: "
 
+	githubAPITokenWarning = "WARNING: Github has API rate limits for all requests which do not contain a token.\nPlease create a read-only access token to increase that limit. https://github.com/settings/tokens"
```
preslavmihaylov, line 23: Could you instead link https://developer.github.com/v3/#rate-limiting to this warning message.

preslavmihaylov: This link shows details about why one might want to consider adding an apitoken

authmanager/authmanager.go
```diff
@@ -19,21 +20,31 @@ const (
...
+			return acquireAPIToken(cfg)
+		case config.AuthTypeOffline:
+			return acquireOfflineToken(cfg.Auth)
+		default:
+			panic("unknown auth token type")
+		}
+	}()
+
+	if cfg.Auth.Token == "" {
+		showAuthWarning(cfg)
```
preslavmihaylov, line 44: This code should be in [config.Validate](https://github.com/preslavmihaylov/todocheck/blob/ce4a2420e46ddfc5795358363e05399d874a9b4f/config/config.go#L145) because that's the place where we do validations of the user's configuration.

Apart from that, there is no need to change the rest of the `AcquireToken` code, more specifically, the error handling. I don't see how wrapping this in a function helps.

authmanager/authmanager.go
```diff
@@ -110,3 +121,12 @@ func extractBaseURL(origin string) string {
 
 	return fmt.Sprintf("%s//%s", tokens[0], tokens[1])
 }
+
+func showAuthWarning(cfg *config.Local) {
+	switch cfg.IssueTracker {
+	case config.IssueTrackerGithub:
+		fmt.Println(color.RedString(githubAPITokenWarning))
```
preslavmihaylov, line 128: This is fine, but please use a yellow color for the warning instead of red. We use red for errors and reusing it for warnings is confusing.

preslavmihaylov (changes requested):
Thanks for preparing this PR!

I have several code quality comments, after which we are ready to ship this.

authmanager/authmanager.go
codykaup: Good idea. That's where I started but figured it might be better to do it after we acquire the user's token (which also stores it if the user supplies a token in the CLI). Now that I think about it, it makes more sense to do it up front during validation.

config/config.go
```diff
@@ -181,6 +188,15 @@ func (l *Local) validateAuthOfflineURL() error {
 	return nil
 }
 
+func (l *Local) showAuthWarning() {
+	switch l.IssueTracker {
+	case IssueTrackerGithub:
```
preslavmihaylov, line 193: Could you move this condition for the issue tracker being github in the if statement?
ie:
```
if l.IssueTracker == IssueTrackerGithub && l.Auth.Token == "" {
    fmt.Fprintln(color.Output, color.YellowString(githubAPITokenWarning))
}
```

This simplifies the changes to 3 LOC. Apart from that making the check in `showAuthWarning` is misleading because:
```
if l.Auth.Token == "" {
	l.showAuthWarning()
}
```

looks like the warning is always printed if the auth token is missing, which is not correct. The reader has to look at the `showAuthWarning` details to understand the whole picture.

The change with `color.Output` is done because otherwise, the colored output won't render properly in a windows terminal.

preslavmihaylov (changes requested):
Thanks for making the changes. It's a lot simpler now.

I have one minor comment regarding readability & one major one regarding making this feature compatible with Windows.

After those are done, we are ready to merge.

config/config.go
```diff
@@ -25,6 +26,8 @@ const (
 	IssueTrackerGitlab               = "GITLAB"
 	IssueTrackerPivotal              = "PIVOTAL_TRACKER"
 	IssueTrackerRedmine              = "REDMINE"
+
+	githubAPITokenWarning = "WARNING: Github has API rate limits for all requests which do not contain a token.\nPlease create a read-only access token to increase that limit.\nGo to https://developer.github.com/v3/#rate-limiting for more information."
```
preslavmihaylov, line 30: Could you run your version of the app locally & provide a screenshot of how this warning looks like?

It would be nice to see the end result being rendered.

config/config.go
codykaup: Gotcha, that makes sense! I initially wanted it to be expandable but it makes more sense at this point to focus in on Github like you mentioned.

config/config.go
codykaup: Absolutely!

Screen Shot 2020-10-17 at 9 26 38 AM

preslavmihaylov (approved):
Great! Thanks for taking the time to get this through.

Merging it to master.

codykaup:
> Great! Thanks for taking the time to get this through.
> 
> Merging it to master.

No, no. Thanks for working with me on it!!

PR #73, opened 2020-10-16
extract the config.Validate function into a separate package

preslavmihaylov:
This refactor would allow future changes to the `Validate` function to include dependencies on `issuetracker` and others which is not currently possible as that would lead to cyclic dependencies.


preslavmihaylov:
Abandoned due to tedious git rebasing issues...

PR #76, opened 2020-10-18
Show error if auth type is incorrect

zen37:
closes #issue-71

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,53 +3,20 @@ package config
 import (
 	"fmt"
 	"io/ioutil"
-	"net/url"
 	"os"
 	"regexp"
 	"strings"
 
-	"github.com/fatih/color"
 	"gopkg.in/yaml.v2"
 )
 
 // DefaultLocal contains the default filepath to the local todocheck config for the current repository
 const DefaultLocal = ".todocheck.yaml"
 
-// IssueTracker enum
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -181,6 +185,16 @@ func (l *Local) validateAuthOfflineURL() error {
 	return nil
 }
 
+func (l *Local) validateAuthType() error {
+	for _, authType := range AuthTypes {
+		if l.Auth.Type == authType {
+			return nil
+		}
```
preslavmihaylov, line 192: Could you follow a similar approach to [this](https://github.com/preslavmihaylov/todocheck/pull/74/files#diff-b20fa56bdafc4db95f805c14326690bf70c3e4932928c846923fa83caa24b9beR35) for consistency's sake?

ie, having an `IsValid` function on the `AuthType`

config/config.go
```diff
@@ -160,6 +160,10 @@ func (l *Local) Validate() []error {
 		}
 	}
 
+	if err := l.validateAuthType(); err != nil {
```
preslavmihaylov, line 163: pull latest master, there were changes in this file.

preslavmihaylov (changes requested):
This is good. Thanks for taking the time to pull this off. I have some minor comments, after which we can merge this.

PR #77, opened 2020-10-19
Add ExpectJSONOutput

### LINKED ISSUE #23: Add tests for JSON output format
Currently, todocheck supports json output, but there are no integration tests in-place to cover this use-case.

Extend the `testing/scenariobuilder` to support a `ExpectJSONOutput` function. if that is specified, the `scenariobuilder` should expect the output to be in JSON format, rather than the standard one.

To use `todocheck` with json output, use the `--format json` flag.

The json array elements order shouldn't matter.

Copy one of the existing scenarios & add the `ExpectJSONOutput` function to the test scenario.

szliao:
closes #23

### DIFF
--- a/testing/scenariobuilder/builder.go
+++ b/testing/scenariobuilder/builder.go
@@ -35,6 +35,7 @@ type TodocheckScenario struct {
 	versionFlagRequested   bool
 	deleteTokensCacheAfter bool
 	expectedExitCode       int
+	expectJsonFormat       bool
 	issueTracker           issuetracker.Type
 	issues                 map[string]issuetracker.Status
 	envVariables           map[string]string
@@ -174,6 +175,12 @@ func (s *TodocheckScenario) ExpectExecutionError() *TodocheckScenario {
 	return s
 }
 
+// ExpectJsonOutput sets the output to be in JSON format
+func (s *TodocheckScenario) ExpectJsonOutput() *TodocheckScenario {
+	s.expectJsonFormat = true
+	return s
+}
+
 // Run sets up the environment & executes the configured scenario
 func (s *TodocheckScenario) Run() error {
 	var err error
@@ -185,6 +192,12 @@ func (s *TodocheckScenario) Run() error {
 	cmd := exec.Command(s.binaryLoc, "--basepath", s.basepath, "--config", s.cfgPath)
 	if s.versionFlagRequested {
 		cmd.Args = append(cmd.Args, "--version")
+	}
+
+	format := ""
+	if s.expectJsonFormat {
+		format = "json"
+		cmd.Args = append(cmd.Args, "--format", format)
 	}
 
 	teardown, err := s.setupTestEnvironment(cmd)
@@ -211,8 +224,15 @@ func (s *TodocheckScenario) Run() error {
 		}
 	}
 
+	var actualOutput string
+	if format == "json" {
+		actualOutput = stdout.String()
+	} else {
+		actualOutput = stderr.String()
+	}
+
 	validateFuncs := []validateFunc{
-		validateTodoErrs(stderr.String(), s.todoErrScenarios),
+		validateTodoErrs(actualOutput, s.todoErrScenarios, format),
 		validateAuthTokensCache(s.cfg.Auth.TokensCache, s.cfg.Auth.OfflineURL, s.expectedAuthToken),
 	}
 
--- a/testing/scenariobuilder/builder.go
+++ b/testing/scenariobuilder/builder.go
@@ -175,8 +175,8 @@ func (s *TodocheckScenario) ExpectExecutionError() *TodocheckScenario {
 	return s
 }
 
-// ExpectJsonOutput sets the output to be in JSON format
-func (s *TodocheckScenario) ExpectJsonOutput() *TodocheckScenario {
+// WithJSONOutput sets the output to be in JSON format
+func (s *TodocheckScenario) WithJSONOutput() *TodocheckScenario {
 	s.expectJsonFormat = true
 	return s
 }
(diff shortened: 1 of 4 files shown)


testing/scenariobuilder/builder.go
```diff
@@ -174,6 +175,12 @@ func (s *TodocheckScenario) ExpectExecutionError() *TodocheckScenario {
 	return s
 }
 
+// ExpectJsonOutput sets the output to be in JSON format
+func (s *TodocheckScenario) ExpectJsonOutput() *TodocheckScenario {
```
preslavmihaylov, line 179: I think this should be `WithJSONOutput` instead (sorry, I know I suggested the name initially :) ), because it changes the input of the program AND the expected output.

testing/scenariobuilder/validations.go
```diff
@@ -12,17 +13,38 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-func validateTodoErrs(programOutput string, scenarios []*TodoErrScenario) validateFunc {
+func validateTodoErrs(programOutput string, scenarios []*TodoErrScenario, format string) validateFunc {
```
preslavmihaylov, line 16: I feel like it would be harder to maintain this function moving forward if we stick to this approach with instrumenting `if format == "json"` wherever its needed.

Instead, let's make two functions `validateStandardTodoErrs` and `validateJSONTodoErrs`.
In this function, you'd do:
```go
if format == "json" {
    return validateJSONTodoErrs(programOutput, scenarios)
} else {
    return validateStandardTodoErrs(programOutput, scenarios)
}
```

preslavmihaylov: This will enable you to validate todo errors as its most suitable to do so for each format.

Currently, when validating json TODO errors, you are trying to fit the json-specific validation with the standard validation and it's quite unnatural.

By extracting these two functions, you will be able to validate both formats in the best suitable way for each.

testing/scenariobuilder/validations.go
```diff
@@ -65,10 +87,21 @@ func validateAuthTokensCache(tokensCache string, url, expectedToken string) vali
...
-			return i
+		if format == "json" {
+			bytes, err := scenarios[i].ToJSON()
+			if err != nil {
+				panic(err)
+			}
+
+			if string(bytes) == target {
+				return i
+			}
```
preslavmihaylov, line 100: When you do the suggested refactoring of the functions, you can validate the json scenarios by simply marshalling all scenarios to JSON and comparing it against the program output.

preslavmihaylov (changes requested):
Thanks a lot for putting this together @szliao!

I have some comments regarding improving readability & maintainability of the new code, but in general, from a user's/test-creator's perspective, it is exactly what I'd expect!

preslavmihaylov:
Also, apologies for the late review, I see you've submitted this 5 days ago. The previous week was a bit busy for me, but I've got more time throughout this weekend and next week, so I will provide more timely feedback.

szliao:
@preslavmihaylov 
Thank you for your advice. Also sorry for late response&fix. I've fixed them as you suggested. Can you check it again?

preslavmihaylov (changes requested):
Thanks, this is a lot better than the previous version. :)

I know only have a comment on the specific way the new function is implemented. After that is resolved, we are good to go

testing/scenariobuilder/validations.go
```diff
@@ -37,6 +39,56 @@ func validateTodoErrs(programOutput string, scenarios []*TodoErrScenario) valida
...
+		for _, elem := range elements {
+			element := elem
+			actualByte, err := json.Marshal(&element)
+			if err != nil {
+				return err
+			}
+
+			removeIdx := -1
+			for j, expected := range scenarioBytes {
+				if bytes.Equal(actualByte, expected) {
```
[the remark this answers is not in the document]
szliao: The way you suggested looks better. I've made a change. Can you check it again?

testing/todocheck_test.go
```diff
@@ -72,6 +72,26 @@ func TestFirstlineMalformedTodo(t *testing.T) {
...
+		WithConfig("./test_configs/no_issue_tracker.yaml").
+		WithJSONOutput().
+		ExpectTodoErr(
+			scenariobuilder.NewTodoErr().
+				WithType(errors.TODOErrTypeMalformed).
+				WithLocation("scenarios/firstline_comment/main.cpp", 1)).
+		ExpectTodoErr(
+			scenariobuilder.NewTodoErr().
+				WithType(errors.TODOErrTypeMalformed).
+				WithLocation("scenarios/firstline_comment/other.cpp", 1)).
```
preslavmihaylov, line 88: Could you also add a scenario for `TODOErrTypeNonExistent` and `TODOErrTypeClosed`.
This will guarantee that the JSON output is valid for those cases as well.

You can reuse the Jira localhost mock server, which is provided by the testing framework.
See the test `TestAnnotatedTodos` for an example.

preslavmihaylov (changes requested):

testing/todocheck_test.go
szliao: I added a new test function to test scenario for `TODOErrTypeNonExistent` and `TODOErrTypeClosed`.
https://github.com/preslavmihaylov/todocheck/pull/77/files#diff-4d213ea2bbf5c6daa9454def85352fba6ebdfe5860d3578dfc060a389bc15ce4R167

preslavmihaylov (approved):
Thanks for this! Now it's good to go

preslavmihaylov:
I'm fine with these changes now but it seems you will have to rebase first

PR #81, opened 2020-10-24
refactored issue tracker component, extracting specific issue tracker code into separate components

preslavmihaylov:
part 1 of #82

I've made this refactoring to enable a more easier integration of new issue trackers.

The current implementation is quite rigid as it is based on various `switch-cases`, scattered throughout the whole project.
One should have a very deep understanding of the project to be able to create a new integration with an issue tracker.

With this change, all issue tracker-related functions are encapsulated in separate packages, which are independent from the rest of the system and the higher-level interface.

This is part 1 of the refactoring. There are still other things to do, but I will do them with a separate PR:
 * Add the `issuetracker.HealthcheckURL` and related functions into the `issutracker.IssueTracker` interface
 * Extract the authentication middleware functions into the new package as well

## Test plan
I've tested the tool works without issues with all supported issue trackers (github, gitlab, redmine, pivotal, jira).

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -14,19 +14,18 @@ import (
 
 // Fetcher for task statuses by contacting task management web apps' rest api
 type Fetcher struct {
-	origin         string
-	tracker        config.IssueTracker
+	issuetracker.IssueTracker
 	authMiddleware authmiddleware.Func
 }
 
 // NewFetcher instance
-func NewFetcher(origin string, tracker config.IssueTracker, authMw authmiddleware.Func) *Fetcher {
-	return &Fetcher{origin, tracker, authMw}
+func NewFetcher(issueTracker issuetracker.IssueTracker, authMw authmiddleware.Func) *Fetcher {
+	return &Fetcher{issueTracker, authMw}
 }
 
 // Fetch a task's status based on task ID
 func (f *Fetcher) Fetch(taskID string) (taskstatus.TaskStatus, error) {
-	req, err := http.NewRequest("GET", f.origin+issuetracker.TaskURLSuffixFor(taskID, f.tracker), nil)
+	req, err := http.NewRequest("GET", f.IssueTracker.IssueAPIOrigin()+f.IssueTracker.TaskURLFrom(taskID), nil)
 	if err != nil {
 		return taskstatus.None, fmt.Errorf("failed creating new GET request: %w", err)
 	}
@@ -49,7 +48,7 @@ func (f *Fetcher) Fetch(taskID string) (taskstatus.TaskStatus, error) {
 		return taskstatus.None, fmt.Errorf("bad status code upon fetching task: %d - %s", resp.StatusCode, string(body))
 	}
 
-	task := issuetracker.TaskFor(f.tracker)
+	task := f.IssueTracker.TaskModel()
 	err = json.Unmarshal(body, &task)
 	if err != nil {
 		return taskstatus.None, fmt.Errorf("couldn't unmarshal response task JSON: %w", err)
--- a/issuetracker/github/github.go
+++ b/issuetracker/github/github.go
@@ -20,6 +20,10 @@ func (it *IssueTracker) TaskModel() issuetracker.Task {
 
 // TaskURLFrom taskID returns the url for the target github task ID to fetch
 func (it *IssueTracker) TaskURLFrom(taskID string) string {
+	if len(taskID) > 0 && taskID[0] == '#' {
+		return taskID[1:]
+	}
+
 	return taskID
 }
 
--- a/issuetracker/github/github.go
+++ b/issuetracker/github/github.go
@@ -1,0 +1,35 @@
+package github
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/preslavmihaylov/todocheck/common"
+	"github.com/preslavmihaylov/todocheck/issuetracker"
+)
+
+// IssueTracker implementation for integrating with public & private github issue trackers
+type IssueTracker struct {
+	Origin string
+}
+
+// TaskModel returns the model representing a deserialized github task
+func (it *IssueTracker) TaskModel() issuetracker.Task {
+	return &Task{}
+}
+
+// TaskURLFrom taskID returns the url for the target github task ID to fetch
+func (it *IssueTracker) TaskURLFrom(taskID string) string {
+	return taskID
+}
+
+// IssueAPIOrigin returns the URL for github's issue-fetching API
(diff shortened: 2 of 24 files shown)


mehdy:
Just checking in to say this is a big one :)) So It may take a little longer than usual.

preslavmihaylov:
Yeah, I agree. Please take your time. Let's not break anything. :)

fetcher/fetcher.go
```diff
@@ -14,19 +14,18 @@ import (
...
-func NewFetcher(origin string, tracker config.IssueTracker, authMw authmiddleware.Func) *Fetcher {
-	return &Fetcher{origin, tracker, authMw}
+func NewFetcher(issueTracker issuetracker.IssueTracker, authMw authmiddleware.Func) *Fetcher {
+	return &Fetcher{issueTracker, authMw}
 }
 
 // Fetch a task's status based on task ID
 func (f *Fetcher) Fetch(taskID string) (taskstatus.TaskStatus, error) {
-	req, err := http.NewRequest("GET", f.origin+issuetracker.TaskURLSuffixFor(taskID, f.tracker), nil)
+	req, err := http.NewRequest("GET", f.IssueTracker.IssueAPIOrigin()+f.IssueTracker.TaskURLFrom(taskID), nil)
```
mehdy, line 28: I think it would be better if `IssueTracker` provided an abstract method and then here we didn't need to concatenate these two.

issuetracker/github/github.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// IssueAPIOrigin returns the URL for github's issue-fetching API
+func (it *IssueTracker) IssueAPIOrigin() string {
+	tokens := common.RemoveEmptyTokens(strings.Split(it.Origin, "/"))
+	if !strings.HasPrefix(tokens[0], "http") {
+		tokens = append([]string{"https:"}, tokens...)
+	}
+
+	scheme, owner, repo := tokens[0], tokens[2], tokens[3]
```
mehdy, line 33: I always try to avoid using direct access with indexes since it's so easy to face an `index out of range` error and hard to maintain in the future.
Instead, we can use `regexp` for example.

I suggest moving the part of extracting data from `it.Origin` to a private method as well.

issuetracker/gitlab/gitlab.go
```diff
@@ -0,0 +1,37 @@
...
+
+// TaskURLFrom taskID returns the url for the target gitlab task ID to fetch
+func (it *IssueTracker) TaskURLFrom(taskID string) string {
+	return taskID
+}
+
+// IssueAPIOrigin returns the URL for github's issue-fetching API
+func (it *IssueTracker) IssueAPIOrigin() string {
+	tokens := common.RemoveEmptyTokens(strings.Split(it.Origin, "/"))
+	if tokens[0] != "http:" && tokens[0] != "https:" {
```
mehdy, line 30: The use of `HasPrefix` method (from the GitHub implementation) is much cleaner I think.

issuetracker/pivotaltracker/pivotaltracker.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// TaskModel returns the model representing a deserialized redmine task
+func (it *IssueTracker) TaskModel() issuetracker.Task {
+	return &Task{}
+}
+
+// TaskURLFrom taskID returns the url for the target redmine task ID to fetch
+func (it *IssueTracker) TaskURLFrom(taskID string) string {
+	if len(taskID) > 0 && taskID[0] == '#' {
```
mehdy, line 23: `strings.HasPrefix` can make this cleaner as well.

issuetracker/pivotaltracker/pivotaltracker.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/preslavmihaylov/todocheck/common"
+	"github.com/preslavmihaylov/todocheck/issuetracker"
+)
+
+// IssueTracker implementation for integrating with public & private redmine issue trackers
```
mehdy, line 11: All docstrings in this file mention `redmine` instead of `pivotal tracker`.

mehdy (changes requested):
There are other instances of the issues I raised.

issuetracker/github/github.go
preslavmihaylov: I agree that using `regexp` would be much better here. To avoid making too many changes in a single PR, would you agree to file an issue for doing that instead?

As for extracting it to a private function, I think that would be unnecessary given that the function is ~8 LOC long. If we do that, then this public function will simply become a wrapper. What do you think?

issuetracker/gitlab/gitlab.go
preslavmihaylov: I agree. I didn't know of this function. Similarly to my comment above, would you agree to add this in scope for a follow-up task to avoid making too many changes in this PR?

fetcher/fetcher.go
preslavmihaylov: You're right, I agree this is a good idea.

Does `IssueURLFor(taskID string)` sound like a good name?

Additionally, I also suggest doing this in a follow-up PR to avoid too many changes from refactoring the original approach.

issuetracker/pivotaltracker/pivotaltracker.go
preslavmihaylov: True, I missed this one.

I changed it in the latest commit.

fetcher/fetcher.go
mehdy: Sure! the name is good enough.

So I'll resolve this one since it'll be addressed in another PR.

issuetracker/github/github.go
mehdy: Of course, let's go with an issue for later. I like to keep the PRs small an simple as well.

About the private method, although I don't consider a public method which is only a wrapper for some other private methods a bad practice, I think we can stick we this for now since it's tiny as you mentioned.

issuetracker/gitlab/gitlab.go
mehdy: Sure!

mehdy:
@preslavmihaylov I think this is ready to be merged if you're done with it.

preslavmihaylov:
Yeah, I'm done with it. I've also described the follow-up tasks [here](https://github.com/preslavmihaylov/todocheck/issues/82#issuecomment-720043278)

Let me know if I missed something

mehdy (approved):
This is good to go!

Issue #78, opened 2020-10-24
Validation for existing github repository always fails if the repository is private


preslavmihaylov:
In #70, a new validation was added which first checks that a repository exists before proceeding with the rest of the application.

For private github repositories, this always fails because the `apitoken` is not acquired before making the validation.
What's more, the API token is not added as an authorisation middleware when checking that the repository exists.

What to do about it:
 * Acquire Authentication token before making any validations
 * Instrument calls to `IsHealthy` in `fetcher` to plug in the authorisation middleware

preslavmihaylov:
This issue also makes this warning:
![image](https://user-images.githubusercontent.com/7642330/97077369-d762b400-15eb-11eb-954f-cfe046430ede.png)

always visible as it checks for the existence of the token, without acquiring it first

preslavmihaylov:
No need to create a new release after this is fixed as these changes are done after the latest release (#68)

preslavmihaylov:
I've mitigated this incident, but I had to remove the #70 validation. In a follow-up PR, I will introduce it again.

PR #85, opened 2020-10-25
unblock issues caused by bad validation

preslavmihaylov:
Part 1 for #78 

 * Acquire the auth token before the validation begins to avoid displaying the github warning every time
 * Remove the issue tracker exists validation as it always fails for private github repositories

The fix I've made here addresses the issue with the warning always present and partially, the one with always failing validation for private github repositories.

What's left is to also include the authentication middleware in calls to `fetcher.IsHealthy`

I will do that after finishing the refactorings started in #81.
For the time being, I've removed that validation to unblock master from not working with private github repositories.

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -27,10 +27,6 @@ func Validate(cfg *config.Local) []error {
 		errs = append(errs, err)
 	}
 
-	// if err := validateIssueTrackerExists(cfg); err != nil {
-	// 	errs = append(errs, err)
-	// }
-
 	if err := validateAuthType(cfg); err != nil {
 		errs = append(errs, err)
 	}
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -27,9 +27,9 @@ func Validate(cfg *config.Local) []error {
 		errs = append(errs, err)
 	}
 
-	if err := validateIssueTrackerExists(cfg); err != nil {
-		errs = append(errs, err)
-	}
+	// if err := validateIssueTrackerExists(cfg); err != nil {
+	// 	errs = append(errs, err)
+	// }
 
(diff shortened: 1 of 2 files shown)


validation/validation.go
```diff
@@ -27,9 +27,9 @@ func Validate(cfg *config.Local) []error {
 		errs = append(errs, err)
 	}
 
-	if err := validateIssueTrackerExists(cfg); err != nil {
```
mehdy, line 30: Shouldn't this work correctly when the token is provided?
And if not, I think removing the code completely without commenting it out could be a better approach. I think we should leverage git features for this sort of works.

preslavmihaylov: > Shouldn't this work correctly when the token is provided?
Yes, it should. The problem is that it currently doesn't take the token into consideration when making the HTTP request to check for the existence of the repository.

That's what I aim to address in a follow-up PR.

> And if not, I think removing the code completely without commenting it out could be a better approach. I think we should leverage git features for this sort of works.

What do you mean? How could we leverage git features to check for the existence of the repository?

mehdy: > Yes, it should. The problem is that it currently doesn't take the token into consideration when making the HTTP request to check for the existence of the repository.
> 
> That's what I aim to address in a follow-up PR.
Got it. It's alright then.

> What do you mean? How could we leverage git features to check for the existence of the repository?

No No. I meant **commenting out** the code is a bad practice and instead we should rely on git history for codes we do not need right now but maybe someday in the future.
So I suggest removing these lines and adding them again when we actually need it.

preslavmihaylov: Alright, makes sense

mehdy (approved):
It's good to go!

Issue #71, opened 2020-10-13
Show an error if auth type is incorrect


preslavmihaylov:
Here's the error I get when the provided auth type in `.todocheck.yaml` is incorrect:
![image](https://user-images.githubusercontent.com/7642330/95826513-174bb080-0d3b-11eb-9ddd-04ebf85f56f1.png)

This happens because there is a check for it in one of the components using the config. It would be more appropriate to validate against this in the `config.Validate` function instead and show an error message to the user in a normal fashion, rather than relying on a panic, which happens afterwards.

akk5597:
Would the solution be to call `authmanager.AcquireToken` in `config.Validate`, and replace the `panic` with an `error` saying that the auth token type is unknown?

preslavmihaylov:
I think a good approach would be to make a function like this (taken from [here](https://www.ribice.ba/golang-enums/):
```
func (lt LeaveType) IsValid() error {
    switch lt {
    case AnnualLeave, Sick, BankHoliday, Other:
        return nil
    }
    return errors.New("Invalid leave type")
}
```

This function should be in the same file as the enum declarations to avoid someone creating a new enum in the future without upgrading this func.

Using `AcquireToken` to validate is not a good idea, IMO, because it is not meant to be used for validation, but rather, for acquiring a token as the name states.

zen37:
> It would be more appropriate to validate against this in the config.Validate function instead 

Where is the config.Validate function? I could not find it.

preslavmihaylov:
https://github.com/preslavmihaylov/todocheck/blob/ce4a2420e46ddfc5795358363e05399d874a9b4f/config/config.go#L145

zen37:
sorry, I didn't know of this convention, fileName.functionName and it is logical.

preslavmihaylov:
Closed by #84

PR #87, opened 2020-10-28
re-fixes issue #71 broken by PR #85

zen37:
Application panicked if an incorrect auth type was in `.todocheck.yaml`. 
It was fixed but then PR #85 made the change to acquire the auth token before the validation begins to avoid displaying the github warning every time. This way an incorrect auth type was no longer validated. 
The fix was to put the auth type validation in the acquire token function.

### DIFF
--- a/authmanager/authmanager.go
+++ b/authmanager/authmanager.go
@@ -37,7 +37,7 @@ func AcquireToken(cfg *config.Local) error {
 	case config.AuthTypeOffline:
 		return acquireOfflineToken(cfg.Auth)
 	default:
-		panic("this shouldn't have happened, don't know what to do here")
+		panic("It's on us! Don't know how to handle this authentication token type. Please file an issue here - https://github.com/preslavmihaylov/todocheck/issues/new")
 	}
 }
 
--- a/authmanager/authmanager.go
+++ b/authmanager/authmanager.go
@@ -24,6 +24,11 @@ const (
 
 // AcquireToken stores the issue tracker's auth token based on the auth type specified
 func AcquireToken(cfg *config.Local) error {
+
+	if !cfg.Auth.Type.IsValid() {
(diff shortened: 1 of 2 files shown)


authmanager/authmanager.go
```diff
@@ -24,6 +24,11 @@ const (
 
 // AcquireToken stores the issue tracker's auth token based on the auth type specified
 func AcquireToken(cfg *config.Local) error {
+
+	if !cfg.Auth.Type.IsValid() {
+		return fmt.Errorf("invalid auth type: %q. valid auth types are: %q", cfg.Auth.Type, config.ValidAuthTypes)
+	}
```
preslavmihaylov, line 30: Could you also remove this validation from `validation.Validate`?

There's no value in having in in two places.

authmanager/authmanager.go
```diff
@@ -32,7 +37,7 @@ func AcquireToken(cfg *config.Local) error {
 	case config.AuthTypeOffline:
 		return acquireOfflineToken(cfg.Auth)
 	default:
-		panic("unknown auth token type")
+		panic("this shouldn't have happened, don't know what to do here")
```
preslavmihaylov, line 40: Could you change this to:
```
It's on us! Don't know how to handle this authentication token type. Please file an issue here - https://github.com/preslavmihaylov/todocheck/issues/new
```

preslavmihaylov (changes requested):
Thanks for your contribution!

A few final touches & it's good to go

preslavmihaylov:
Could you also add a description, explaining what exactly the problem was from #71 and how it got broken in #85?

I don't think it will be obvious to someone without the context we have

preslavmihaylov (approved):
This is now good to go. Thanks for the contribution! :)

PR #89, opened 2020-10-29
Issue #75: extract TestValidOrigins and TestInvalidOrigins to unit test

### LINKED ISSUE #75: Extract TestValidOrigins and TestInvalidOrigins to a unit test instead of system test
Currently, in `testing/todocheck_test.go`, there are two tests which are actually unit tests, while the tests in this file are meant to be system tests - testing the todocheck binary in a black box manner.

The tests are:
 * `TestValidOrigins` 
 * `TestInvalidOrigins`

What to do about it?
 * Extract those two tests into `validation/validation_test.go`.
 * Add a new `make` step for running all unit tests in the source code. `go test` probably?
 * Add a new github action in `.github/workflows/master.yaml` for triggering the unit tests on all PRs.

RenanOD:
Closes #75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ build:
 
 test: build
 	go test -v -count=1 ./testing
+	go test -v -count=1 ./validation
 
 release:
 	@echo "Generating binaries for version $(version)..."
--- a/testing/todocheck_test.go
+++ b/testing/todocheck_test.go
@@ -4,10 +4,8 @@ import (
 	"testing"
 
 	"github.com/preslavmihaylov/todocheck/checker/errors"
-	"github.com/preslavmihaylov/todocheck/config"
 	"github.com/preslavmihaylov/todocheck/testing/scenariobuilder"
 	"github.com/preslavmihaylov/todocheck/testing/scenariobuilder/issuetracker"
-	"github.com/preslavmihaylov/todocheck/validation"
 )
 
 func TestSingleLineMalformedTodos(t *testing.T) {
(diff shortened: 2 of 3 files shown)


testing/todocheck_test.go
```diff
@@ -405,65 +403,6 @@ func TestInvalidGithubAccess(t *testing.T) {
 	}
 }
 
-func TestInvalidOrigins(t *testing.T) {
-	invalidConfigPaths := []string{
-		"./test_configs/invalid_github_https.yaml",
```
RenanOD, line 410: Noticed that I changed the path to avoid moving the `.yaml` files.

validation/validation_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"testing"
+
+	"github.com/preslavmihaylov/todocheck/validation"
+	"github.com/preslavmihaylov/todocheck/config"
+)
+
+func TestInvalidOrigins(t *testing.T) {
+	invalidConfigPaths := []string{
+		"../testing/test_configs/invalid_github_https.yaml",
```
RenanOD, line 12: Noticed that I changed the path to avoid moving the .yaml files. Maybe they should be moved?

Makefile
```diff
@@ -6,6 +6,7 @@ build:
 
 test: build
 	go test -v -count=1 ./testing
+	go test -v -count=1 ./validation
```
preslavmihaylov, line 9: Can you make this step invoke testing for all tests excluding `./testing`?

This way, if someone adds a new unit test in the future, they won't have to go through this exercise again.

Additionally, you can remove the `-count=1` option as it's only relevant for system tests.

Also, could you please separate this in a new step - e.g. `unit_test`. This way, if someone sees his CI step failing, he will know whether it's from the unit tests or integration tests.

Finally, could you modify `.github/workflows/master.yaml` to use this new make step to invoke all unit tests in CI?

validation/validation_test.go
```diff
@@ -0,0 +1,67 @@
...
+		"../testing/test_configs/invalid_redmine_origin.yaml",
+		"../testing/test_configs/invalid_redmine_port.yaml",
+	}
+
+	for _, path := range invalidConfigPaths {
+		cfg, err := config.NewLocal(path, ".")
+		if err != nil {
+			t.Errorf("%s", err)
+			continue
+		}
```
preslavmihaylov, line 31: Using `config.NewLocal` is not a good idea given that we're writing a unit test in the `validation` package. Here, the tests should only test the `validation` package, instead of also testing the rest.

Instead, what you can do is:
```
cfg := config.Local{
  Origin: <your-test-origin>,
  IssueTracker: <your-test-issue-tracker>
}

errors := validation.Validate(cfg)
...
```

This way, you are testing the `validation package in isolation and what's more, you avoid having to create all those extra files.

So, to sum this up, we'll have to create a test case for each of those files (not necessarily in separate functions) and delete those files if they're no longer used in the `testing` directory.

validation/validation_test.go
```diff
@@ -0,0 +1,67 @@
...
+		"../testing/test_configs/valid_gitlab_origin.yaml",
+		"../testing/test_configs/valid_gitlab_port.yaml",
+		"../testing/test_configs/valid_gitlab_subdomain.yaml",
+		"../testing/test_configs/valid_jira_origin.yaml",
+		"../testing/test_configs/valid_jira_port.yaml",
+		"../testing/test_configs/valid_jira_subdomain.yaml",
+		"../testing/test_configs/valid_pivotal_origin.yaml",
+		"../testing/test_configs/valid_redmine_origin.yaml",
+		"../testing/test_configs/valid_redmine_port.yaml",
+		"../testing/test_configs/valid_redmine_subdomain.yaml",
```
preslavmihaylov, line 53: Same goes for here.

preslavmihaylov (changes requested):
Thanks for taking this on!

It will, however, need some additional refactoring to transform this from a semi-system test to an actual unit test.

See the comments I've left for details.

preslavmihaylov:
Seems like this was a hacktoberfest rush and is abandoned. Closing.

PR #98, opened 2020-11-07
Added flag verbose and passed its value deeper by #95

### LINKED ISSUE #95: Too much verbosity
If project have many dot-files and ignored paths then each run followed by showing many lines of ignored files `Skipping ignored file...`.  It will be nice to make the todocheck execution was quiet. And add option `-verbose`/`-v` for show all ignored files and may be other lines on future. And use option -v to verbose output instead of version.

sattellite:
Added flag `--verbose` with default `false` value. Its value passed to Traverser constructor and deeper to comments Traverser.

Maybe it is necessary to extend the displayed lines when the verbose is enabled? For example, all processed files or all files where TODO strings are found? 

Closed #95

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -5,7 +5,7 @@ import (
 	"sync"
 )
 
-var once sync.Once
+var lock sync.Mutex
 var log *Logger
 
 // Logger provided simple logger.
@@ -15,11 +15,14 @@ type Logger struct {
 
 // Setup configures logger
 func Setup(verbose bool) {
-	once.Do(func() {
+	lock.Lock()
+	defer lock.Unlock()
+
+	if log == nil {
 		log = &Logger{
 			verbose,
 		}
-	})
+	}
 }
 
 // Info prints in verbose mode to standart output. Format according to fmt.Println.
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -14,7 +14,7 @@ type Logger struct {
 }
 
 // Setup configures logger
-func Setup(verbose bool) *Logger {
+func Setup(verbose bool) {
 	lock.Lock()
 	defer lock.Unlock()
 
@@ -23,7 +23,6 @@ func Setup(verbose bool) *Logger {
 			verbose,
 		}
 	}
-	return log
 }
 
 // Info prints in verbose mode to standart output. Format according to fmt.Println.
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -18,10 +18,8 @@ func Setup(verbose bool) {
 	lock.Lock()
 	defer lock.Unlock()
 
-	if log == nil {
-		log = &Logger{
-			verbose,
-		}
+	log = &Logger{
+		verbose,
 	}
 }
 
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -26,25 +26,25 @@ func Setup(verbose bool) {
 }
 
 // Info prints in verbose mode to standart output. Format according to fmt.Println.
-func (l *Logger) Info(args ...interface{}) {
-	Info(args...)
+func Info(args ...interface{}) {
+	log.Info(args...)
 }
 
 // Info prints in verbose mode to standart output. Format according to fmt.Println.
-func Info(args ...interface{}) {
-	if log.verbose {
+func (l *Logger) Info(args ...interface{}) {
+	if l.verbose {
 		fmt.Println(args...)
 	}
 }
 
 // Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
-func (l *Logger) Infof(format string, args ...interface{}) {
-	Infof(format, args...)
+func Infof(format string, args ...interface{}) {
+	log.Infof(format, args...)
 }
 
 // Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
-func Infof(format string, args ...interface{}) {
-	if log.verbose {
+func (l *Logger) Infof(format string, args ...interface{}) {
+	if l.verbose {
 		fmt.Printf(format, args...)
 	}
 }
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,0 +1,42 @@
+package logger
+
+import "fmt"
+
+var log *Logger
+
+// Logger provided simple logger.
+type Logger struct {
+	verbose bool
+}
+
+// Setup configures logger
+func Setup(verbose bool) *Logger {
+	log = &Logger{
+		verbose,
+	}
+	return log
+}
+
+// Info prints in verbose mode to standart output. Format according to fmt.Println.
+func (l *Logger) Info(args ...interface{}) {
+	Info(args...)
+}
+
+// Info prints in verbose mode to standart output. Format according to fmt.Println.
+func Info(args ...interface{}) {
+	if log.verbose {
+		fmt.Println(args...)
+	}
+}
+
+// Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
+func (l *Logger) Infof(format string, args ...interface{}) {
+	Infof(format, args...)
+}
+
+// Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
+func Infof(format string, args ...interface{}) {
+	if log.verbose {
+		fmt.Printf(format, args...)
+	}
+}
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,7 +1,11 @@
 package logger
 
-import "fmt"
+import (
+	"fmt"
+	"sync"
+)
 
+var lock sync.Mutex
 var log *Logger
 
 // Logger provided simple logger.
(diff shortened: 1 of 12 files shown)


preslavmihaylov (changes requested):
I think passing a `verbose` flag around any component which needs it is not a scalable way to have this feature.

Instead, I suggest creating a new `logger` component which exposes one static function for now:
 * `logger.Info`

Then, the usage in other components would be:
```go
logger.info("skipping ignored file ...")
```

For usability's sake, it would be good for this function to accept formatting parameters.

Finally, in `main.go` the `verboseRequested` flag should be passed to e.g. `logger.Setup(...)` which configures the global logger.

And finally, in the `logger.Info` function, the contents should only be printed if verbose mode is requested.

main.go
```diff
@@ -28,6 +28,7 @@ func main() {
 	var basepath = flag.String("basepath", ".", "The path for the project to todocheck. Defaults to current directory")
 	var cfgPath = flag.String("config", "", "The project configuration file to use. Will use the one from the basepath if not specified")
 	var format = flag.String("format", "standard", "The output format to use. Available formats - standard, json")
+	var verboseRequested = flag.Bool("verbose", false, "Verbose mode")
```
preslavmihaylov, line 31: Let's change the description to:
```
Make todocheck more talkative
```

similar to how curl defines their verbose flag:
```
$ curl --help | grep verbose
  --trace-time    Add time stamps to trace/verbose output
  -v, --verbose       Make the operation more talkative

```

sattellite:
> which configures the global logger

Does this mean, that in `main.go` `log` variable will be in global scope? Global scope is bad practice.

I prepared two variants of code. Prefix is used for show that it is the same instance:
1. Logger creates self instance in self scope and return itself with special method. In `main.go` used `logger.Setup(...)` and in other componetns uses `logger.GetInstance()`. GetInstance and Setup can be merged together in one complex method. https://play.golang.org/p/XBtUw5hwJUz
2. Logger creates in main.go and passes to ither components. Classic and simple variant. https://play.golang.org/p/Xu00rAfMLqb

What do you prefer?

preslavmihaylov:
It is true that global scope is bad practice but it all depends on the context. Nothing is black or white :) 

In this specific scenario, I think that using a global context is fine because the global component isn't providing an interface worth mocking or testing against, neither is that interface going to lead to side-effects which will affect how the user packages will function. 

The value it brings, however, outweighs the drawback I mentioned because you won't have to provide an explicit dependency in all components which need to log something.

An example of using global scope in Go's standard library is the `http` package, which maintains a global http client. The benefits of having that is that it greatly simplifies using the library for simple use-cases because you can directly invoke `http.Get` without initialising any objects or configuring anything beforehand.

If one has more advanced needs, they can explicitly initialize an http client and configure it as needed.

sattellite:
I have created more simple variant for use. User can call instance methods and package functions. 

```go
import ".../logger"

func main() {
	// Initialize logger instance and call its methods 
	l := logger.Setup(true, "#>")
	l.Info("Method call")
	
	// Or just call package functions. It will use what been passed to Setup.
	// Setup can be called once anywhere. First Setup was called upper, when created instance
	logger.Info("Function call")
}
```

https://play.golang.org/p/maQxNdYp-eB

```go
import (
	"fmt"
)

var logger *Logger

type Logger struct {
	verbose bool
	prefix  string
}

func Setup(verbose bool, prefix string) *Logger {
	logger = &Logger{
		verbose,
		prefix,
	}
	return logger
}

func (l *Logger) Info(args ...interface{}) {
	Info(args...)
}

func Info(args ...interface{}) {
	if logger.verbose {
		fmt.Print(logger.prefix)
		fmt.Println(args...)
	}
}
```

sattellite:
I agree with nothing is black or white. But it is very strange. My inner perfectionist resists such a decision. Maybe we should explicitly import the logger component before usage? In last third variant it will be just imported and ready to use.

logger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+
+var log *Logger
+
+// Logger provided simple logger.
+type Logger struct {
+	verbose bool
+}
+
+// Setup configures logger
+func Setup(verbose bool) *Logger {
```
preslavmihaylov, line 13: I don't think it's a good idea for this function to return the global instance of the logger as users might change the configuration of the instance they've received, which would, as a side-effect change the behavior of the global logger.

Instead, I suggest making this function void - ie, not returning any instance. And if, in the future, the need arises, we can introduce a `NewLogger` function which explicitly creates an instance of a logger, which can be configured as needed.

logger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+	log = &Logger{
+		verbose,
+	}
```
preslavmihaylov, line 16: Let's make this initialization thread-safe just in case we decide to introduce concurrency in the tool in the future.

This can be achieved using a `sync.Mutex` similar to [this gist](https://gist.github.com/jeffotoni/8bc07d67519751296cee98c4ff1c311a#file-golang-singleton-goroutine-new-2-go)

No need for the singleton pattern, just the explicit locking.

For more info, check out [this article](https://medium.com/golang-issue/how-singleton-pattern-works-with-golang-2fdd61cd5a7f)

logger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+// Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
+func (l *Logger) Infof(format string, args ...interface{}) {
+	Infof(format, args...)
+}
+
+// Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
+func Infof(format string, args ...interface{}) {
+	if log.verbose {
+		fmt.Printf(format, args...)
+	}
```
preslavmihaylov, line 41: This should be:
```
log.Infof(format, args...)
```

logger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+// Info prints in verbose mode to standart output. Format according to fmt.Println.
+func Info(args ...interface{}) {
+	if log.verbose {
+		fmt.Println(args...)
+	}
+}
+
+// Infof formatted prints in verbose mode to standart output. Format according to fmt.Printf.
+func (l *Logger) Infof(format string, args ...interface{}) {
+	Infof(format, args...)
```
preslavmihaylov, line 34: This should be:
```go
if l.verbose {
    fmt.Printf(format, args...)
}
```

Otherwise, you are always using the global logger regardless of the fact that you've initialized your own instance

logger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+// Info prints in verbose mode to standart output. Format according to fmt.Println.
+func (l *Logger) Info(args ...interface{}) {
+	Info(args...)
+}
+
+// Info prints in verbose mode to standart output. Format according to fmt.Println.
+func Info(args ...interface{}) {
+	if log.verbose {
+		fmt.Println(args...)
+	}
```
preslavmihaylov, line 29: Same here

logger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+func Setup(verbose bool) *Logger {
+	log = &Logger{
+		verbose,
+	}
+	return log
+}
+
+// Info prints in verbose mode to standart output. Format according to fmt.Println.
+func (l *Logger) Info(args ...interface{}) {
+	Info(args...)
```
preslavmihaylov, line 22: Same here

preslavmihaylov (changes requested):

preslavmihaylov:
> Maybe we should explicitly import the logger component before usage? In last third variant it will be just imported and ready to use.

I'm not sure I fully understand what you're saying but looking at your changes in `traverser/lines/lines.go`, that is exactly the usage I feel like is best.

logger/logger.go
sattellite: Thank you for pointing to new problems and patterns. I am newbie in golang and all new knowledge is important for me.

sattellite:
> I'm not sure I fully understand what you're saying

I have little experience in English. Сode more accurately describes my thoughts

logger/logger.go
```diff
@@ -0,0 +1,47 @@
...
+	verbose bool
+}
+
+// Setup configures logger
+func Setup(verbose bool) {
+	once.Do(func() {
+		log = &Logger{
+			verbose,
+		}
+	})
```
preslavmihaylov, line 22: Instead of using `sync.Once`, we should use `sync.Mutex` as otherwise, this function will initialize the global logger only the first time it is invoked, which is not the desired behavior

preslavmihaylov (changes requested):

logger/logger.go
sattellite: Reverted

preslavmihaylov: but I see that you're still using the `if log == nil` which essentially does the same thing as `sync.Once`.

My point was to not make the logger a singleton, as it is now, but rather to just guard its initialization via a mutex.

The short version is, we don't need the if clause

sattellite: I thought we needed a singleton here. And manage it with method like SetVerose. Ok, will fix it.

preslavmihaylov (approved):

PR #99, opened 2020-11-13
Add integration tests for github public/private repositories

### LINKED ISSUE #79: Create a test for testing integration with private & public github repositories
This test should only run on CI and the API token should be passed via the `TODOCHECK_AUTH_TOKEN` environment variable

preslavmihaylov:
Closes #79 

In order to avoid regressions in integrations with supported issue trackers (Pivotal, Redmine, Github, Gitlab, etc) and avoid the need for extensive manual testing before accepting any PR, I've started preparing a test suite which verifies that todocheck can correctly integrate with all supported issue trackers for both private and public repositories.

In this PR, I'm creating the test suite for integrating with private & public github repositories. To do that, I have created a github secret which contains the api key for reading the repositories and is available to the tests in the github action.

Additionally, I had to fix an issue where the `setupMockIssueTrackerServer` in the scenario builder was always setting up a mock server regardless if one was needed. This lead to some small regressions in the rest of the integration tests due to an invalid issue tracker origin as per our validations.

Finally, these tests have been setup to only run in CI (github actions) because they rely on a sensitive secret which is available in the CI environment thanks to Github secrets, but is not available in the local environment of any of the maintainers.


mehdy (approved):
LGTM!

Issue #53, opened 2020-09-04
Add support for custom TODO format per project


preslavmihaylov:
Create a new section in the `.todocheck.yaml` config which enables one to specify a per-project custom TODO format.

E.g.
```
custom_todos:
 - FIXME
 - @todo
 - @fixme
```

This will allow users to integrate todocheck with other kinds of linters which demand a more specific todo format.

In case one specifies a TODO which matches one of this patterns but not its casing (e.g. `FiXmE`), that TODO should be detected as invalid.

Include test scenarios for this new use case as well as the edge cases mentioned above.

Finally, document the new feature in the `README` or create a follow-up task for that.

preslavmihaylov:
Mentioned in #48

preslavmihaylov:
@addiks FYI, this feature is now live.

Feel free to pull latest master & `go build` or `go install github.com/preslavmihaylov/todocheck` 🎉

PR #100, opened 2020-11-13
Add support for custom TODO format

sattellite:
This PR related to #53


preslavmihaylov:
Looks good so far. 

Thanks for taking this on, I think it will be a bumpy issue to execute, but very valuable and much appreciated if you manage to pull it off!

Some pointers - you can look at the `matchers` package and make the regexes in e.g. `matchers/scripts/todomatcher.go` to accept the TODO as a parameter.

To avoid making too many if-else statements, meant to preserve the current behaviour, you can make the config accept a default value for `custom_todos` as `[]string{"TODO"}` and always rely on this field. That way, you won't have to make checks such as "if this field is present, use behaviour A, otherwise, use behaviour B"

preslavmihaylov:
Also, the default `TODO` format should always be supported, regardless of the field being present or not.
Additionally, I would love to see some integration tests in the `testing` package in the end to see this in action.

config/config.go
```diff
@@ -154,3 +157,19 @@ func removeCurrentDirReference(dirs []string) {
 func isRelativePath(path string) bool {
 	return path[0] != '/' && path[0] != '~' && !windowsAbsolutePathPattern.MatchString(path)
 }
+
+// checkTodos trying find default TODO string and adding it if not exists
+func checkTodos(todos []string) []string {
```
preslavmihaylov, line 162: The name `checkTodos` doesn't really review the func's intent. One has to read the docstring or implementation to understand its purpose.

I suggest renaming this to something more understandable such as `addDefaultFormatIfMissing`

sattellite: Renamed

sattellite:
I only have to add tests for this functionality.

preslavmihaylov:
thanks @sattellite I'll spend some time to review this during the weekend as it's a rather hard week for me so far. :(

matchers/matchers.go
```diff
@@ -147,3 +161,10 @@ func SupportedFileExtensions() []string {
 
 	return exts
 }
+
+func prepareTodosForRegex(todos []string) string {
+	if len(todos) == 0 {
+		panic("Empty list of todo strings")
+	}
+	return fmt.Sprintf("(?:%s)", strings.Join(todos, "|"))
+}
```
preslavmihaylov, line 170: This is not the place for such a function.

The problem is that you are writing it with the knowledge that all underlying todo matchers operate with regex.
This is an implementation detail and if at some point, we decide to migrate from regex to something else, this function here will cause a lot of issues.

Instead, this function can be defined in the `common` package as `arrayAsRegexAnyMatchExpression`.
The function declaration should instead be `[]string` and the specifics of turning the array of strings into a regex expression can be done for every todo matcher.

matchers/php/todomatcher.go
```diff
@@ -2,37 +2,83 @@ package php
...
+var singleLineTodoPattern *regexp.Regexp
+var singleLineValidTodoPattern *regexp.Regexp
 
-var multiLineTodoPattern = regexp.MustCompile("(?s)^\\s*/\\*.*TODO")
-var multiLineValidTodoPattern = regexp.MustCompile("(?s)^\\s*/\\*.*TODO (#?[a-zA-Z0-9\\-]+):.*")
+var singleLineScriptTodoPattern *regexp.Regexp
+var singleLineScriptValidTodoPattern *regexp.Regexp
+
+var multiLineTodoPattern *regexp.Regexp
+var multiLineValidTodoPattern *regexp.Regexp
```
preslavmihaylov, line 19: I guess these global variables are used so that we don't initialise all these regexps every time a new matcher for a file is created, which happens for every file.

Instead of making this optimization here, you can do it in the matcher factories. There, instead of initializing a matcher every time, initialise them as fields of the `matcherFactory` and reuse the instances.

If we implement it with these global variables, it would lead to one having to understanding the implementation details of this component to use them properly.

Given the folllowing snippet:
```
matcherA = php.NewTodomatcher("TODO", "todo")
matcherB = php.NewTodomatcher("FIXME", "fixme")
```

`matcherA` will now work with `FIXME` and `fixme` instead of the formats it was initialised with.

matchers/php/todomatcher.go
```diff
@@ -2,37 +2,83 @@ package php
...
 
 // NewTodoMatcher for php comments
-func NewTodoMatcher() *TodoMatcher { return &TodoMatcher{} }
+func NewTodoMatcher(todos string) *TodoMatcher {
+	lock.Lock()
+	defer lock.Unlock()
```
preslavmihaylov, line 24: There is no need for using this once we remove the global variables.

matchers/php/todomatcher.go
```diff
@@ -2,37 +2,83 @@ package php
...
+		singleLineScriptValidTodoPattern = regexp.MustCompile("^\\s*# " + todos + " (#?[a-zA-Z0-9\\-]+):.*")
+	}
+
+	// Multiline line
+	if multiLineTodoPattern == nil {
+		multiLineTodoPattern = regexp.MustCompile("(?s)^\\s*/\\*.*" + todos)
+	}
+	if multiLineValidTodoPattern == nil {
+		multiLineValidTodoPattern = regexp.MustCompile("(?s)^\\s*/\\*.*" + todos + " (#?[a-zA-Z0-9\\-]+):.*")
+	}
```
preslavmihaylov, line 48: Once we remove the global variables, all these can be condensed into a single return statement, which creates the struct.

matchers/python/todomatcher.go
```diff
@@ -2,30 +2,64 @@ package python
```
preslavmihaylov, line 2: The comment on the previous todo matcher applies for all of them.

Not going to repeat the same stuff all over the place.

testing/scenarios/custom_todos/standart/main.go
```diff
@@ -0,0 +1,29 @@
+package main
```
preslavmihaylov, line 1: The directory should be `standard` not `standart`

matchers/matchers.go
```diff
@@ -118,11 +120,23 @@ var supportedMatchers = map[string]*matcherFactory{
...
+type Matchers struct {
+	todos []string
+}
+
+// NewMatchers returns Matcher instance
+func NewMatchers(customTodos []string) *Matchers {
+	return &Matchers{
+		todos: customTodos,
+	}
+}
```
preslavmihaylov, line 133: Instead of creating a `Matchers` struct here just to hold the custom todos, you can add an additional field in the `checker` and pass along the custom todos on each call to `TodoMatcherForFile `

preslavmihaylov (changes requested):
I haven't commented on absolutely everything, but there's plenty to work on still.

After the next iteration of changes, I'll go over the PR again with additional comments.

However, you're moving in the right direction.

config/config.go
```diff
@@ -154,3 +157,25 @@ func removeCurrentDirReference(dirs []string) {
...
+	for i, v := range todos {
+		if v == "TODO" {
+			isExists = true
+			continue
+		}
+		// Remove escaping "\" for "@". YAML can't start scalar from @.
+		// "@" in YAML is reserved indicator https://yaml.org/spec/1.2/spec.html#id2772075
+		if strings.HasPrefix(v, "\\@") {
+			todos[i] = v[1:]
+		}
```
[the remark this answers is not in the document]
sattellite: Added function to decode escaped strings. I want add tests for it but I don't understand where to place and how to name the  tests for it.

testing/scenarios/custom_todos/standart/main.go
sattellite: Renamed the directory.

matchers/php/todomatcher.go
sattellite: I don't understand what you mean with this words. I have never seen or done this before. Can you give me links for read about this? Or describe little more please. This is my first language with strict types and I have never worked with factories before. So I don't understand how to transfer this optimization from matchers to the factory. I would be very grateful for your help.

sattellite: > Instead of making this optimization here, you can do it in the matcher factories. There, instead of initializing a matcher every time, initialise them as fields of the `matcherFactory` and reuse the instances.

I thought for all night about the problem described and I think I understood what you wrote about. Is it a correct solution?

```go
var lock sync.Mutex

type matcherFactory struct {
	newTodoMatcher     func(string) TodoMatcher
	newCommentsMatcher func(callback state.CommentCallback) CommentMatcher
	todoInstance       TodoMatcher
}

var standardMatcherFactory = &matcherFactory{
	func(customTodos string) TodoMatcher {
		if standardMatcherFactory.todoInstance == nil {
			lock.Lock()
			standardMatcherFactory.todoInstance = standard.NewTodoMatcher(customTodos)
			lock.Unlock()
		}
		return standardMatcherFactory.todoInstance
	},
	func(callback state.CommentCallback) CommentMatcher {
		return standard.NewCommentMatcher(callback, false)
	},
}
```

matchers/matchers.go
sattellite: Fixed

matchers/matchers.go
sattellite: Renamed and moved function to `common`

matchers/php/todomatcher.go
preslavmihaylov: yes, like this but no need for the lazy initialization.

Simply initialise it like so:
```go
var standardMatcherFactory = &matcherFactory{
        todoMatcher: standard.NewTodoMatcher()
	func(customTodos string) TodoMatcher {
		return standardMatcherFactory.todoMatcher
	},
	func(callback state.CommentCallback) CommentMatcher {
		...
	},
}
```

The `customTodos` need not be provided here. Provide them in the instance functions of the todo matcher. This avoids creating a state variable.

sattellite: Sorry in advance for stupid questions and inferences. I guess I don't understand what it should look like at all.

> The `customTodos` need not be provided here. Provide them in the instance functions of the todo matcher. This avoids creating a state variable.

Is this mean that I need add function like `setTodos` for matchers? When this function must be called?

I don't understand why this change will make the code clearer. There will be state variables with regular expressions in any matcher. This will make it possible to change sets of regular expressions at runtime. The configuration is set once at startup and is not changed at runtime, so I don't see the point in this method.

preslavmihaylov: hmm, okay. You're right that the matchers will have the regexps as state variables & if we provide the `customTodos` on each call to `IsMatch` and `IsValid`, then the regular expressions will need to be reinitialized.

In this case, this approach is ok with a suggestion for improving the synchronization like this:
```go
var (
	once            *sync.Once
	standardMatcher TodoMatcher
)

var (
	standardMatcherFactory = &matcherFactory{
		func(customTodos string) TodoMatcher {
			once.Do(func() {
				standardMatcher = standard.NewTodoMatcher(customTodos)
			})

			return standardMatcher
		},
		func(callback state.CommentCallback) CommentMatcher {
			return standard.NewCommentMatcher(callback, false)
		},
	}
        ...
```

sattellite: Optimized. Thank you for your advice.

testing/todocheck_test.go
```diff
@@ -802,3 +802,644 @@ func TestPrintingVersionFlagStopsProgram(t *testing.T) {
 		t.Errorf("%s", err)
 	}
 }
+
+func TestScriptsCustomTodos(t *testing.T) {
```
preslavmihaylov, line 806: Could you extract all new tests from this file into a `customtodos_test.go`?

Just to keep files small

traverser/todoerrs/todoerrs.go
```diff
@@ -15,9 +15,9 @@ import (
 type TodoErrCallback func(todoerr *errors.TODO) error
 
 // NewTraverser for todo errors
-func NewTraverser(f *fetcher.Fetcher, ignoredPaths []string, callback TodoErrCallback) *Traverser {
+func NewTraverser(f *fetcher.Fetcher, ignoredPaths, todos []string, callback TodoErrCallback) *Traverser {
 	return &Traverser{
-		comments.NewTraverser(ignoredPaths, commentsCallback(checker.New(f), callback)),
+		comments.NewTraverser(ignoredPaths, commentsCallback(checker.New(f, todos), callback)),
```
preslavmihaylov, line 20: I feel like `todos` is ambiguous here, especially given that it isn't an explicit type, it is a `[]string`.

Suggest renaming it to `customTodos` for clarity

preslavmihaylov (changes requested):
This is almost complete now.

I have several minor issues to address still, after which we are finally ready to merge.

Thanks for all your efforts, this was a big one!

matchers/matchers.go
```diff
@@ -28,54 +29,85 @@ type CommentMatcher interface {
...
+
 var (
 	standardMatcherFactory = &matcherFactory{
-		func() TodoMatcher {
-			return standard.NewTodoMatcher()
+		func(customTodos []string) TodoMatcher {
+			once.Do(func() {
+				if standardMatcher == nil {
+					standardMatcher = standard.NewTodoMatcher(customTodos)
+				}
```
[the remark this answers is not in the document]
sattellite: `standardMatcher` used in two places: `standardMatcherFactory` and `standardMatcherWithNestedMultilineCommentsFactory`. This check will avoid second initialization of `standardMatcher`

sattellite:
Please check README. English is not my native language and I often make mistakes in it.

preslavmihaylov (changes requested):

matchers/matchers.go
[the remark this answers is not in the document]
sattellite: You are was right. I wrote test for multiple matchers and it was failed. I rewrote code with your advice with changed `once` scope and memoized `matcher` and test passed.

matchers/matchers.go
```diff
@@ -33,82 +33,93 @@ type matcherFactory struct {
...
-		func(customTodos []string) TodoMatcher {
-			once.Do(func() {
-				scriptsMatcher = scripts.NewTodoMatcher(customTodos)
-			})
-			return scriptsMatcher
-		},
+		func() func([]string) TodoMatcher {
+			return func(customTodos []string) TodoMatcher {
+				var once sync.Once
+				var matcher TodoMatcher
```
preslavmihaylov, line 71: Why are these inside the returned function instead of being in the closure, similar to how you've done it for `standard` matchers?

preslavmihaylov (changes requested):

matchers/matchers.go
sattellite: Apparently, I was distracted. I corrected this misprint.

README.md
```diff
@@ -242,10 +243,22 @@ ignored:
 Ignored files/folders can be specified via standard pattern-matching.
 Hidden files (dotfiles, i.e. `.git`, `.gitignore`, etc) are ignored by default.
 
+# Custom Todos
+You can customize your set of variables to be considered as todos. Specify it in the `custom_todos` section in the `.todocheck.yaml`:
```
preslavmihaylov, line 247: Let's rephrase it a bit to make the target use-case more evident:
> By default, `todocheck` looks for todos in the format `// TODO #231: ...`. Most projects stick to this format. 
> 
> However, if you have some special todo format, which is different from the default one, you can specify a > list of custom todo formats in your `.todocheck.yaml`

README.md
```diff
@@ -242,10 +243,22 @@ ignored:
...
+```
+origin: some.origin.com
+issue_tracker: JIRA
+custom_todos:
+  - "@fix"
+  - ToDo
+```
+
+Note that no matter what list is specified a default "TODO" will always be added to it.
```
preslavmihaylov, line 256: > Note that this list adds additional formats the tool should match, but doesn't substitute the default one - `TODO`.

README.md
```diff
@@ -242,10 +243,22 @@ ignored:
...
+# Custom Todos
+You can customize your set of variables to be considered as todos. Specify it in the `custom_todos` section in the `.todocheck.yaml`:
+```
+origin: some.origin.com
+issue_tracker: JIRA
+custom_todos:
+  - "@fix"
+  - ToDo
+```
+
```
preslavmihaylov, line 255: Let's also specify an example usage:

With the above configuration, `todocheck` will not start matching todo comments in the following formats:
```
// @fix #234: a leftover task...
// ToDo #15: A funky-looking todo comment
```

preslavmihaylov (changes requested):

preslavmihaylov:
Wow, we've come a long way with this PR. Thank you tremendously for your efforts!

Really looking forward to shipping this soon.

sattellite:
Thank you for spending so much time checking my code. You are a great software architect. And you are a very cool reviewer. This pull request was very useful for me thanks to your experience and efforts.

preslavmihaylov (approved):
I am very happy to finally approve these changes & hit the merge button!

Congratulations on such a significant contribution 🎉

PR #119, opened 2020-11-29
add ci tests for gitlab integration

preslavmihaylov:
Related to #120 

One of the CI steps fail because the new secret is not available for pull_request_target runs. As observed, when the event is `push`, the step passes. After this is merged, the failing step will, hence, start passing.

In addition to this, I've decided to reduce the test verbosity to more easily identify which tests are failing.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -3,6 +3,8 @@ name: master
 on:
   push:
   pull_request_target:
+    branches:
+      - master
 
 jobs:
   build:
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -3,8 +3,6 @@ name: master
 on:
   push:
   pull_request_target:
-    branches:
-      - master
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


mehdy:
@preslavmihaylov The tests are still failing.

preslavmihaylov:
> @preslavmihaylov The tests are still failing.

Yeah, I've mentioned that in the description - tests whichare run with `pull_request_target` run with the workflow as-is from the master branch. The workflow from there doesn't have the new secret I've added.

After the diff is merged, the test will start passing, as is seen by the `push` variant of the step.

mehdy:
> > @preslavmihaylov The tests are still failing.
> 
> Yeah, I've mentioned that in the description - tests whichare run with `pull_request_target` run with the workflow as-is from the master branch. The workflow from there doesn't have the new secret I've added.
> 
> After the diff is merged, the test will start passing, as is seen by the `push` variant of the step.

Sorry. I missed the comment. :))

mehdy (approved):
Everything looks good to me.

PR #122, opened 2020-11-30
add pivotaltracker CI tests

preslavmihaylov:
Similarly to #119 and #99 I'm creating CI tests for pivotaltracker.

For this purpose, I've created an account on their site with read-only access to a pivotaltracker project with a set of hardcoded issues with all various statuses.

For each of them, I've created a separate todo line in the tests in this PR to verify we can correctly unmarshal the response received from their API.

I also changed my mind about removing the verbosity of tests as making them non-verbose turned out to make it quite hard to see where the issue is due to mised standard output from tests and the testing output

The `pull_request_target` tests are failing as they are using the workflow from the master branch, which doesn't include the new secret. This is expected.


mehdy (approved):
LGTM!

mehdy:
@preslavmihaylov Would you please rebase it with master?

PR #127, opened 2021-01-07
Migrated 'is repo exists' logic into issuetracker component

### LINKED ISSUE #75: Extract TestValidOrigins and TestInvalidOrigins to a unit test instead of system test

preslavmihaylov:
I've extended the `issuetracker.IssueTracker` interface with another routine, which is inherently connected to the issue tracker - checking if the tracker exists, previously named `healthcheck`.

However, to support this feature fully, we need to first migrate the `authMiddleware` logic from `authmanager/authmiddleware` into the issuetracker components as well as otherwise, the new `Exists` routine won't work for private instances of issue trackers, which require an api token.

For now, all calls to `Exists` return `true` but at least the function resides in the appropriate package(s).

Another unrelated change, which was necessary after doing this migration was to move the `TestValidOrigin` and `TestInvalidOrigin` from the `testing` package into the `validation` package. This had to be done at some point either ways (#75)

Closes #75


preslavmihaylov:
@mehdy don't be intimidated by the amount of files changed. Most of them are moving fixtures from `testing/test_configs` into `validation/fixtures`

mehdy:
@preslavmihaylov Sure! :)) Thanks for the heads up :D

mehdy (approved):
All seems good to me.

Issue #82, opened 2020-10-24
Refactor the issue tracker related components/functions


preslavmihaylov:
The current implementation is quite rigid as it is based on various switch-cases, scattered throughout the whole project.
One should have a very deep understanding of the project to be able to create a new integration with an issue tracker.

The scope of this task is to refactor the current issuetracker component along with all other components which have issuetracker-related code (ie `fetcher.IsHealthy`, `authmiddleware`) by encapsulating all functions one has to write when integrating a new issuetracker in a single `issuetracker.IssueTracker` interface.

The specific issue tracker implementations should reside in separate packages (eg `issuetracker/github`, `issuetracker/jira`, etc) and they should all implement that interface.

By making this change, making integrations with issue trackers in the future will be much easier as all code one has to write is confined to a single package.

preslavmihaylov:
Follow-up tasks after the review of #81:
 * Use `regexp` instead of `strings.Split` for getting the values from an issuetracker origin
 * Use `strings.HasPrefix` instead of checking a substring manually in `IssueAPIOrigin` and `TaskURLFrom` for all issue trackers
 * Combine `IssueAPIOrigin` and `TaskURLFrom` into a single function as they are used in a single spot and always concatenated

mehdy:
@preslavmihaylov Yesterday, I opened these issues (#90 #91 #92) for them as well :)) If you want to keep track of them in this issue, feel free to close them.

preslavmihaylov:
Ah nice, thanks for this! I've leave them in separate issues to encourage collaborators to take them on :)

PR #128, opened 2021-01-08
extracted authentication middleware into issuetracker component

preslavmihaylov:
One part of the `issuetracker` refactoring is to move the authentication middleware logic into the `issuetracker` component.

This would enable easier onboarding of new issue trackers as all routines an issue tracker needs to support are provided by the `issuetracker.IssueTracker` interface. It's much easier this way than having a mental model of the whole codebase to know that there is some sneaky component somewhere, instrumenting an authentication header.

Related to #82


mehdy (approved):
Everything looks good to me.

PR #134, opened 2021-02-23
fixes #121 token passed via env variable is no longer cached

### LINKED ISSUE #121: Don't store auth token if specified via an environment variable
Currently, todocheck provides a way to specify an auth token for a given issue tracker via an environment variable:
```
TODOCHECK_AUTH_TOKEN=<secret> todocheck <args>...
```

This is meant to be used to override the normal flow of prompting for an auth token and persisting it in `~/.todocheck/authtokens.yaml` later.

However, although it overrides the former behavior (prompting for the token), it doesn't override the latter one (persisting it in the tokens file).

Whenever the environment variable `TODOCHECK_AUTH_TOKENS` is specified, the provided token should not be persisted in `~/.todocheck/authtokens.yaml`

sancroth:
Removal of setAndPersistToken call and saving the token in memory.
The call mentioned above would always save the token which is not the required behavior.
The token will now be properly cached on ~/todocheck/authtokens.yaml only when it is set in the config file.

### DIFF
--- a/authmanager/authmanager.go
+++ b/authmanager/authmanager.go
@@ -54,7 +54,9 @@ func acquireToken(authCfg *config.Auth, tokenKey string, instructions string) er
 		authCfg.Token = store.Tokens[tokenKey]
 		return nil
 	} else if envToken := os.Getenv(authTokenEnvVariable); envToken != "" {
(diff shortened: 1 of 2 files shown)


sancroth:
Closed this pr and created an updated one to solve testing errors

PR #135, opened 2021-02-27
fixes#129 offline_url validation

### LINKED ISSUE #129: Validate that offlineURL is set if authentication type is offline
Extend the `validation` package to include a validation that there is an `offline_url` present in the `.todocheck.yaml` configuration if the auth type is `offline`.

sancroth:
Closes #129 

Validate now checks if offline_url is an empty when auth type offline is being used. returns an error is it's not set

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -17,9 +17,9 @@ func Validate(cfg *config.Local, tracker issuetracker.IssueTracker) []error {
 	}
 
 	if cfg.Auth.Type == config.AuthTypeOffline {
-		if err := validateAuthOfflineURL(cfg); err != nil {
+		if err := validateAuthOfflineURLIsSet(cfg); err != nil {
 			errs = append(errs, err)
-		} else if err := validateAuthOfflineURLIsSet(cfg); err != nil {
+		} else if err := validateAuthOfflineURL(cfg); err != nil {
 			errs = append(errs, err)
 		}
 	}
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -19,8 +19,7 @@ func Validate(cfg *config.Local, tracker issuetracker.IssueTracker) []error {
 	if cfg.Auth.Type == config.AuthTypeOffline{
 		if err := validateAuthOfflineURL(cfg); err != nil {
 			errs = append(errs, err)
-		}
-		if err := validateAuthOfflineUrlIsSet(cfg); err != nil {
+		}else if err := validateAuthOfflineURLIsSet(cfg); err != nil {
 			errs = append(errs,err)
 		}
 	}
@@ -60,7 +59,7 @@ func validateAuthOfflineURL(cfg *config.Local) error {
 	return nil
 }
 
-func validateAuthOfflineUrlIsSet(cfg *config.Local) error {
+func validateAuthOfflineURLIsSet(cfg *config.Local) error {
 	if cfg.Auth.OfflineURL==""{
 		return fmt.Errorf("auth type chosen was %q but \"offline_url\" is not set", cfg.Auth.Type)
 	}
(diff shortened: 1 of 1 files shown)


validation/validation.go
```diff
@@ -16,10 +16,16 @@ func Validate(cfg *config.Local, tracker issuetracker.IssueTracker) []error {
...
 
-	if err := validateAuthOfflineURL(cfg); err != nil {
-		errs = append(errs, err)
+	if cfg.Auth.Type == config.AuthTypeOffline{
+		if err := validateAuthOfflineURL(cfg); err != nil {
+			errs = append(errs, err)
+		}
+		if err := validateAuthOfflineUrlIsSet(cfg); err != nil {
+			errs = append(errs,err)
+		}
```
preslavmihaylov, line 25: I'd prefer if we first check that the auth offline url is set and use an `else if` to make the two checks mutually exclusive - if the offline url is not set, there is no point in checking if it's valid.

In example:
```go
if err := validateAuthOfflineUrlIsSet(cfg); err != nil {
   ...
} else if err := validateAuthOfflineURL(cfg); err != nil {
   ...
}
```

validation/validation.go
```diff
@@ -47,13 +53,21 @@ func validateIssueTracker(cfg *config.Local) error {
...
 func validateAuthOfflineURL(cfg *config.Local) error {
-	if _, err := url.ParseRequestURI(cfg.Auth.OfflineURL); cfg.Auth.Type == config.AuthTypeOffline && err != nil {
+	if _, err := url.ParseRequestURI(cfg.Auth.OfflineURL); err != nil {
 		return fmt.Errorf("invalid offline URL: %q", cfg.Auth.OfflineURL)
 	}
 
 	return nil
 }
 
+func validateAuthOfflineUrlIsSet(cfg *config.Local) error {
```
preslavmihaylov, line 63: Please capitalize `URL`. Also, consider setting up `gofmt` as it will probably show a warning in this case.

preslavmihaylov (changes requested):
Thanks for the PR!

Some slight modifications are needed, after which we're good to go.

validation/validation.go
sancroth: Good point! change is now implemented.

validation/validation.go
sancroth: Good point! change is now implemented.

validation/validation.go
```diff
@@ -16,10 +16,15 @@ func Validate(cfg *config.Local, tracker issuetracker.IssueTracker) []error {
 		errs = append(errs, err)
 	}
 
-	if err := validateAuthOfflineURL(cfg); err != nil {
-		errs = append(errs, err)
+	if cfg.Auth.Type == config.AuthTypeOffline{
```
preslavmihaylov, line 19: Could you setup `gofmt` locally so that we avoid the minor issues around formatting such as lacking a space between `AuthTypeOffline` and the curly bracket?

I've also opened a new task to automate this - #138. I think you might find it interesting if you've not dealt with CI setups before :)

preslavmihaylov (changes requested):

validation/validation.go
sancroth: I included gofmt on goland. I used it on visual code but made the change to goland recently and forgot about adding it on the filewatchers. We won't face another issue with this :)

validation/validation.go
preslavmihaylov: One final thing - the checks are still reversed.

It should be `validateAuthOfflineURLIsSet` first and then `validateAuthOfflineURL`. After this, we're good to go.

preslavmihaylov (changes requested):

preslavmihaylov (approved):

PR #136, opened 2021-02-27
fixes #121 token passed via env variable is no longer cached

### LINKED ISSUE #121: Don't store auth token if specified via an environment variable

sancroth:
Removal of setAndPersistToken call and saving the token in memory.
The call mentioned above would always save the token which is not the required behavior.
The token will now be properly cached on ~/todocheck/authtokens.yaml only when it is set in the config file.

Since the TokenCache file will no longer store the token we also need to not run the TokenCache validation test on the scenariobuilder.

### DIFF
--- a/authmanager/authmanager.go
+++ b/authmanager/authmanager.go
@@ -54,7 +54,9 @@ func acquireToken(authCfg *config.Auth, tokenKey string, instructions string) er
 		authCfg.Token = store.Tokens[tokenKey]
 		return nil
 	} else if envToken := os.Getenv(authTokenEnvVariable); envToken != "" {
-		return setAndPersistToken(authCfg, store, tokenKey, envToken)
+		authCfg.Token = envToken
+		store.Tokens[tokenKey] = authCfg.Token
+		return nil
 	}
 
 	fmt.Printf("%s\nToken: ", instructions)
--- a/authmanager/authmanager.go
+++ b/authmanager/authmanager.go
@@ -55,7 +55,6 @@ func acquireToken(authCfg *config.Auth, tokenKey string, instructions string) er
 		return nil
 	} else if envToken := os.Getenv(authTokenEnvVariable); envToken != "" {
 		authCfg.Token = envToken
-		store.Tokens[tokenKey] = authCfg.Token
 		return nil
 	}
 
--- a/testing/scenariobuilder/builder.go
+++ b/testing/scenariobuilder/builder.go
@@ -273,7 +273,13 @@ func (s *TodocheckScenario) Run() error {
 
 	validateFuncs := []validateFunc{
 		validateTodoFunc,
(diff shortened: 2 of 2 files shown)


sancroth:
I managed to solve one the two testing failures that occured after the update on authmanager.go.
Still looking for where the second one occurs.
Now only fails on TestMultipleTodoMatchers test.

sancroth:
The 2 files that are changed should be split into two commits.
The first should apply the changes on testing/scenariobuilder/builder.go which will allow the rest( changes on authmanager/authmanager.go ) to pass the automated tests.

authmanager/authmanager.go
```diff
@@ -54,7 +54,9 @@ func acquireToken(authCfg *config.Auth, tokenKey string, instructions string) er
 		authCfg.Token = store.Tokens[tokenKey]
 		return nil
 	} else if envToken := os.Getenv(authTokenEnvVariable); envToken != "" {
-		return setAndPersistToken(authCfg, store, tokenKey, envToken)
+		authCfg.Token = envToken
+		store.Tokens[tokenKey] = authCfg.Token
```
preslavmihaylov, line 58: You don't need the second line. It will store the token in the tokens cache representation in-memory without persisting it to disk (due to not calling `store.Save`.

testing/scenariobuilder/builder.go
```diff
@@ -273,7 +273,13 @@ func (s *TodocheckScenario) Run() error {
...
 	validateFuncs := []validateFunc{
 		validateTodoFunc,
-		validateAuthTokensCache(s.cfg.Auth.TokensCache, s.cfg.Auth.OfflineURL, s.expectedAuthToken),
+	}
+
+	// only run the TokenCache validation when the environment variable is not set
+	// this test will always fail since the cache is not generated when the
+	// token is set via an environment variable
+	if s.authTokenEnvVariable == "" {
+		validateFuncs = append(validateFuncs,validateAuthTokensCache(s.cfg.Auth.TokensCache, s.cfg.Auth.OfflineURL, s.expectedAuthToken))
```
preslavmihaylov, line 282: For some reason, the tests are still failing. Could you try running the test suite locally & investigating?

You can use `make test` or `go test` directly with a test selector if you don't want to run all tests.

preslavmihaylov (changes requested):
Thanks for the PR!

There are still some issues it appears, but we're almost there.

preslavmihaylov (approved):

PR #137, opened 2021-02-27
fixes #132 authtype allowance validation

### LINKED ISSUE #132: Validate supported authentication types for the supported issuetrackers
Use the issuetracker's `New` functions to make a validation on the authentication type.

If an unsupported authentication type for the given issue tracker is provided, return an error `unsupported authentication type for <issue tracker>: <authentication type>`

Supported authentication types by issue tracker:
Github - none, apitoken
Gitlab - none, apitoken
pivotaltracker - none, apitoken
redmine - none, apitoken
jira - offline

sancroth:
added an issuetracker map that holds the allowed auth types per issuetracker type. if the given auth type does not exist in the perspective array then an error is returned and the issuetracker is not created.

image


sancroth:
Most tests now fail because the yaml config file used is one such as this:
```
origin: http://myjira.com
issue_tracker: JIRA
auth:
  type: none
```

e.g 

```
=== RUN   TestSingleLineMalformedTodos
(standard output follows. Standard output is ignored & not validated...)

(unexpected error occurred. standard error output follows...)
2021/02/27 16:20:11 couldn't create new issue tracker: unsupported authentication type for JIRA: none

    todocheck_test.go:48: program exited with an unexpected exit code: exit status 1
```

preslavmihaylov:
auth type `none` is actually supported for Jira. The description of the ticket is just incomplete.

config/issuetracker.go
```diff
@@ -15,6 +15,14 @@ const (
...
 )
 
+var ValidIssueTrackerAuthTypes = map[IssueTracker][]AuthType{
+	IssueTrackerGithub: { AuthTypeNone , AuthTypeAPIToken},
+	IssueTrackerGitlab: { AuthTypeNone , AuthTypeAPIToken},
+	IssueTrackerPivotal: { AuthTypeNone , AuthTypeAPIToken },
+	IssueTrackerRedmine: { AuthTypeNone , AuthTypeAPIToken },
+	IssueTrackerJira : { AuthTypeOffline },
+}
+
```
preslavmihaylov, line 25: Instead of doing the validation in the `New` functions of each issue tracker, you can add a new function here:
```go
func (it IssueTracker) IsValidAuthType(authType AuthType) bool { ... }
```

Afterwards, you can extend `validation/validation.go` and do the appropriate check there similar to the rest of the code in that file. This will confine the responsibility of validating the configuration in the `validation` package as it is intended.

preslavmihaylov (changes requested):

config/issuetracker.go
sancroth: This was my original thought process but i switched around since the validate method will only run AFTER the issuetracker has been created. I wanted to not create it at all in the first place. I will do it this way then! :)

preslavmihaylov: That's true. It's a bit unfortunate I had to init the issue tracker before the validation, but we can live with this code smell for a while. If you can suggest an alternative which would not demand us to init the issuetracker before the validation, that'd be ideal. :)

sancroth: I noticed that the reason we create the IssueTracker first is to get the perspective token acquisition instructions.
We could create another map per IssueTracker->Message but i saw that some messages are generated with a unique way(using origin or something else on a non standard way) and get the values from the exposed map.

Another way would be to expose a validation.ValidateAuthType method and use it before creating the IssueTracker. The validate package will still handle the validation but there gonna be multiple calls(2 for -now).

validation/validation.go
```diff
@@ -74,3 +78,10 @@ func validateIssueTrackerExists(cfg *config.Local, tracker issuetracker.IssueTra
 
 	return fmt.Errorf("repository %s not found", cfg.Origin)
 }
+
+func validateTrackerAuthType(cfg *config.Local) error {
```
preslavmihaylov, line 82: Could you add unit tests for this new validation in `validation/validation_test.go`?

issuetracker/internal/redmine/redmine.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // New creates a new redmine issuetracker instance
-func New(origin string, authCfg *config.Auth) (*IssueTracker, error) {
+func New(authCfg *config.Auth, origin string) (*IssueTracker, error) {
```
preslavmihaylov, line 14: Why swap the func arguments?

preslavmihaylov (changes requested):

issuetracker/internal/redmine/redmine.go
sancroth: On the first commit `06d7321` the way IssueTracker.New worked needed all 3 arguments to be passed(issuetracker, config, origin) so i used them in the same order as factory.go. After moving the responsibility of the validation to the validation package i removed the IssueTracker but did not revert the order(which is better as origin,cfg to be in par with the struct).

I reverted this again, sorry for causing this confusion :)

preslavmihaylov:
Everything looks good. The only thing left is to add a unit test for the new behavior

preslavmihaylov (approved):
Thanks for the contribution!

preslavmihaylov:
I'll wait for the other PR resolving the broken CI pipeline to get merged and I'll merge this one

Issue #138, opened 2021-02-27
Run gofmt automatically on pull requests


preslavmihaylov:
Extend the github actions to automatically run `gofmt` on every pull request to avoid styling issues resulting from people not having installed gofmt locally.

sancroth:
Please do assign this to me @preslavmihaylov

preslavmihaylov:
For now, we added a CI step which verifies that gofmt is ran locally to format all the necessary files

preslavmihaylov:
It seems it's not as trivial to make gofmt format your code on pull request. For now, we'll suffice with just verifying the format

PR #140, opened 2021-03-01
#138 added a ci step that checks for properly formated golang code

sancroth:
tries to solve #138 This action will check for properly formated code based on gofmt rules.
If any files are not already formated then the build and test steps shall not run and the user will be informed for which files have not been formatted as well as a notice to use gofmt.

The issue that was opened requests for the code to be formatted by the ci itself. That required the code to be pushed back on the pull request source( e.g my fork ) which was causing me errors due to access permissions(access tokens).
Thus i came up with this temporary work around. As i was researching for ways to do this i also found a lot of repos doing something similar and also many people mentioning what i described above.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -7,8 +7,24 @@ on:
       - master
 
 jobs:
+  gofmt-check:
+    name: checking format against gofmt rules
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout new code
+        uses: actions/checkout@v2
+        with: # code from pull request
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+      - name: gofmt test run
+        id: gofmt_run
+        run: |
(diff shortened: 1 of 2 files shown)


preslavmihaylov:
Hmm it seems it wont be as easy to modify the CI pipeline as an external contributor. I will take a look at this and try to achieve the auto formatting this week

sancroth:
I ll check it out a bit more aswell. The available flows i see with this issue are:

a) User makes pull request, format check is done , if it fails then it pushes the format back to the pr which triggers a new test run but formatted this time. This has issues because the cloud instance that runs the actions needs to have write permissions to the branch that the PR comes from.

b) User makes pull request, format check is done, if it fails then it formats and creates a new pull request it self. I don't even know if that's possible and how the user can update with changes later. Doesn't sound like a viable solution at a first glance.

c) Fail the pr if the code included is not already formatted. 

My push tries to implement C.

preslavmihaylov (approved):
I think this suffices for now. It's better than having no check at all. Thanks

PR #143, opened 2021-03-07
fixed wrong directory name that caused ci pipeline failure

### LINKED ISSUE #142: goftm ci check always fails due to wrong directory name
A previous push added the ci-scripts directory that checks if the go code in a pull request is formatted using goftmt.

This breaks the pipeline due to a wrong directory name used in the git workflow.

```
      - name: gofmt test run
        id: gofmt_run
        run: |
          sh ./scripts/gofmt_check.sh
```

sancroth:
fixes #142 The directory name was leftover from local testing on a previous issue. This needs to get through in order for any other pull requests to pass the gofmt check. Sorry for causing this!


preslavmihaylov:
Instead of creating a new `ci-scripts` directory, could you move the existing script into `scripts` instead?

preslavmihaylov (approved):

PR #145, opened 2021-05-22
Add support for .vue files

### LINKED ISSUE #107: Need add support for VueJS files
VueJS have one-file components with file extension `.vue`. One-file components contains html, javascript and css.

JS have standard comments and `.js` is supported by todomatcher: 
```javascript
// oneline comment
/*
multiline comment
*/
```

With preprocessors html and css can have different syntaxes 😞 https://vue-loader.vuejs.org/guide/pre-processors.html 
But it can be parsed as defualt html and css comments:

```html
<!-- HTML oneline and multiline comments -->
```
```css
/* CSS oneline and multiline comments */
```

FelixTheodor:
closes #107 

I added a comment matcher and a todomatcher for .vue-files, supporting oneline JS comments:
`// comment`
as well as multiline JS/CS comments:
`/* comment */`
and HTML comments:
`<!-- comment -->`

I also added some tests for the new file format, and everything seems to work so far.

### DIFF
--- a/matchers/vue/comments.go
+++ b/matchers/vue/comments.go
@@ -27,7 +27,7 @@ type CommentMatcher struct {
 func (m *CommentMatcher) NonCommentState(
 	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
 ) (state.CommentState, error) {
-	if currToken == '/' && prevToken == '/' {
+	if prevToken == '/' && currToken == '/' {
 		m.buffer += string(currToken)
 
 		return state.SingleLineComment, nil
@@ -35,7 +35,7 @@ func (m *CommentMatcher) NonCommentState(
 		m.stringToken = currToken
 
 		return state.String, nil
-	} else if currToken == '*' && prevToken == '/' {
+	} else if prevToken == '/' && currToken == '*'{
 		m.buffer += string(currToken)
 		m.commentType = "CSS"
 
--- a/matchers/vue/comments.go
+++ b/matchers/vue/comments.go
@@ -36,7 +36,9 @@ func (m *CommentMatcher) NonCommentState(
 
 		return state.String, nil
 	} else if prevToken == '/' && currToken == '*' {
-		m.buffer += string(currToken)
+		m.buffer += "/*"
+		m.lines = []string{line}
+		m.linecnt = linecnt
 		m.commentType = "CSS"
 
 		return state.MultiLineComment, nil
@@ -46,6 +48,8 @@ func (m *CommentMatcher) NonCommentState(
 		return state.NonComment, nil
 	} else if m.isStartingHTML && nextToken == '-' {
 		m.buffer += "<!-"
+		m.lines = []string{line}
+		m.linecnt = linecnt
 		m.commentType = "HTML"
 
 		return state.MultiLineComment, nil
@@ -89,19 +93,21 @@ func (m *CommentMatcher) SingleLineCommentState(
 func (m *CommentMatcher) MultiLineCommentState(
 	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
 ) (state.CommentState, error) {
-	if m.isExitingMultilineComment {
-		m.resetState()
-		return state.NonComment, nil
-	}
-
 	m.buffer += string(currToken)
-	if isEndOfMultilineComment(m.commentType, prevToken, currToken, nextToken) {
-		m.buffer += string(nextToken)
+
+	if m.isExitingMultilineComment {
+
 		err := m.callback(m.buffer, filename, m.lines, m.linecnt)
 		if err != nil {
 			return state.NonComment, err
 		}
 
+		m.resetState()
+
+		return state.NonComment, nil
+	}
+
+	if isEndOfMultilineComment(m.commentType, prevToken, currToken, nextToken) {
 		m.isExitingMultilineComment = true
 		return state.MultiLineComment, nil
 	}
--- a/matchers/vue/comments.go
+++ b/matchers/vue/comments.go
@@ -35,7 +35,7 @@ func (m *CommentMatcher) NonCommentState(
 		m.stringToken = currToken
 
 		return state.String, nil
-	} else if prevToken == '/' && currToken == '*'{
+	} else if prevToken == '/' && currToken == '*' {
 		m.buffer += string(currToken)
 		m.commentType = "CSS"
 
--- a/matchers/vue/comments.go
+++ b/matchers/vue/comments.go
@@ -1,0 +1,138 @@
+package vue
+
+import (
+	"github.com/preslavmihaylov/todocheck/matchers/state"
+)
+
+// NewCommentMatcher for vue comments
+func NewCommentMatcher(callback state.CommentCallback) *CommentMatcher {
+	return &CommentMatcher{
+		callback: callback,
+	}
+}
+
+// CommentMatcher for vue comments
+type CommentMatcher struct {
+	callback                  state.CommentCallback
+	buffer                    string
+	lines                     []string
+	linecnt                   int
+	stringToken               rune
+	isExitingMultilineComment bool
+	commentType               string
+	isStartingHTML            bool
+}
+
+// NonCommentState for vue comments
+func (m *CommentMatcher) NonCommentState(
+	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
+) (state.CommentState, error) {
+	if currToken == '/' && prevToken == '/' {
+		m.buffer += string(currToken)
+
+		return state.SingleLineComment, nil
+	} else if currToken == '"' || currToken == '\'' {
+		m.stringToken = currToken
+
+		return state.String, nil
+	} else if currToken == '*' && prevToken == '/' {
+		m.buffer += string(currToken)
+		m.commentType = "CSS"
+
+		return state.MultiLineComment, nil
+	} else if prevToken == '<' && currToken == '!' && nextToken == '-' {
+		m.isStartingHTML = true
+
+		return state.NonComment, nil
+	} else if m.isStartingHTML && nextToken == '-' {
+		m.buffer += "<!-"
+		m.commentType = "HTML"
+
+		return state.MultiLineComment, nil
+	} else if m.isStartingHTML && nextToken != '-' {
+		m.isStartingHTML = false
+	}
+
+	return state.NonComment, nil
+}
+
+// StringState for vue comments
+func (m *CommentMatcher) StringState(
+	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
+) (state.CommentState, error) {
+	if prevToken != '\\' && currToken == m.stringToken {
+		return state.NonComment, nil
+	}
+
+	return state.String, nil
+}
+
+// SingleLineCommentState for vue comments
+func (m *CommentMatcher) SingleLineCommentState(
+	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
+) (state.CommentState, error) {
+	if currToken == '\n' {
+		err := m.callback(m.buffer, filename, []string{line}, linecnt)
+		if err != nil {
+			return state.NonComment, err
+		}
+
+		m.resetState()
+		return state.NonComment, nil
+	}
+
+	m.buffer += string(currToken)
+	return state.SingleLineComment, nil
+}
+
+// MultiLineCommentState for vue comments
+func (m *CommentMatcher) MultiLineCommentState(
+	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
+) (state.CommentState, error) {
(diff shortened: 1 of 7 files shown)


matchers/vue/comments.go
```diff
@@ -0,0 +1,138 @@
...
+
+// NonCommentState for vue comments
+func (m *CommentMatcher) NonCommentState(
+	filename, line string, linecnt int, prevToken, currToken, nextToken rune,
+) (state.CommentState, error) {
+	if currToken == '/' && prevToken == '/' {
+		m.buffer += string(currToken)
+
+		return state.SingleLineComment, nil
+	} else if currToken == '"' || currToken == '\'' {
```
preslavmihaylov, line 34: How is `\` a transition to a string state?

matchers/vue/comments.go
```diff
@@ -0,0 +1,138 @@
...
+		m.stringToken = currToken
+
+		return state.String, nil
+	} else if currToken == '*' && prevToken == '/' {
```
preslavmihaylov, line 38: Please reverse the condition to `prevToken == '/' && currToken == '*'`.

It's easier to read the condition that way. Same applies for the rest of the conditions

matchers/vue/todomatcher.go
```diff
@@ -0,0 +1,65 @@
...
+
+// NewTodoMatcher for vue comments
+func NewTodoMatcher(todos []string) *TodoMatcher {
+	pattern := common.ArrayAsRegexAnyMatchExpression(todos)
+
+	// Single line
+	singleLineTodoPattern := regexp.MustCompile(`^\s*//.*` + pattern)
+	singleLineValidTodoPattern := regexp.MustCompile(`^\s*// ` + pattern + ` (#?[a-zA-Z0-9\-]+):.*`)
+
+	// Multiline line
```
preslavmihaylov, line 18: No need for these comments. They are redundant as the variable names are descriptive enough

matchers/vue/todomatcher.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/preslavmihaylov/todocheck/common"
+	"github.com/preslavmihaylov/todocheck/matchers/errors"
+)
+
+// NewTodoMatcher for vue comments
+func NewTodoMatcher(todos []string) *TodoMatcher {
+	pattern := common.ArrayAsRegexAnyMatchExpression(todos)
+
+	// Single line
```
preslavmihaylov, line 14: No need for these comments. They are redundant as the variable names are descriptive enough

matchers/vue/todomatcher.go
```diff
@@ -0,0 +1,65 @@
...
+type TodoMatcher struct {
+	singleLineTodoPattern      *regexp.Regexp
+	singleLineValidTodoPattern *regexp.Regexp
+	multiLineTodoPattern       *regexp.Regexp
+	multiLineValidTodoPattern  *regexp.Regexp
+}
+
+// IsMatch checks if the current expression matches a vue comment
+func (m *TodoMatcher) IsMatch(expr string) bool {
+	//fmt.Println("trying to find match: ", expr)
```
preslavmihaylov, line 40: leftover debug printing

testing/scenarios/vue/main.vue
```diff
@@ -0,0 +1,16 @@
...
+/* TODO 1: wellformed CS/JS multiline entry */
+/* TODO 2: wellformed CS/JS multiline entry, BUT issue closed */
+/*
+wellformed CS/JS multline entry
+*/
+
+<!--
+TODO 4: wellformed HTML multiline entry
+-->
+
```
preslavmihaylov, line 16: Could you also add an example of a string with a comment inside & verify it's not matched?

e.g.
 * "this is a // valid comment in a string"
 * "this is a /* valid multiline comment in a string */"
 * etc for the rest of the comment types

preslavmihaylov: Additionally, could you add examples for:
 * wellformed multiline HTML comment \w issue closed
 * wellformed single-lien comment \w issue closed
 * malformed multiline CSS comment

preslavmihaylov (changes requested):
This is a very nicely done PR. I have several issues regarding styling & several others regarding additional test scenarios.

Overall, I'm pretty happy with the PR and I'll merge it once you address the changes.

On another note, what did you think about the codebase & how was the experience of extending this feature? Any comments/feedback for improving the codebase?

And how about any project features you feel are important?

FelixTheodor:
Thanks for the very helpful feedback!

I refactored the code as you pointed it out  since it all seem to be very reasonable to me. Gonna push these changes in a minute.

Regarding your general question, the first jump into code was a little confusing for me. I am used to have many little unit-tests written for single methods. This is often very useful when trying to understand and especially debug code for the first time. I couldnt find any of those tests here, so I started to try and write my own little helper tests. Quickly, I noticed that the structure of the code does not "allow" this kind of tests easily since I could not even create a valid call to a method "from scratch". 
Therefore, I just started to write code "blindly" (without instant feedback), looked at the way other file-types are handled and later debugged it with some prints and the integration-test.
Apart from the lack of these unit-tests, I find the code itself very clear and instructive, so it was possible for me to grasp the idea behind most of the things I saw, and reprocude/adapt the concepts to the new problem.

matchers/vue/comments.go
FelixTheodor: It is not, but in this case it functions as a escape sequence to the ' symbol, so the go interpreter know that the literal '-rune is meant and the '-symbol is not the end of the rune sequence. I hope this makes sense to you, and as the new string-tests show, the transition seems to work. I therefore did not change this part of the code.

preslavmihaylov: Oh...

I didn't see that it's a `\'` instead of a `\` only. My bad :)

testing/todocheck_test.go
```diff
@@ -742,6 +742,55 @@ func TestPrintingVersionFlagStopsProgram(t *testing.T) {
...
+				WithLocation("scenarios/vue/main.vue", 7)).
+		ExpectTodoErr(
+			scenariobuilder.NewTodoErr().
+				WithType(errors.TODOErrTypeIssueClosed).
+				WithLocation("scenarios/vue/main.vue", 10)).
+		ExpectTodoErr(
+			scenariobuilder.NewTodoErr().
+				WithType(errors.TODOErrTypeMalformed).
+				WithLocation("scenarios/vue/main.vue", 0).
+				ExpectLine("TODO: malformed CS/JS multline entry, missing number").
```
preslavmihaylov, line 781: Why is the location of this line `0`?

preslavmihaylov: Shouldn't this be:
```go
WithLocation("scenarios/vue/main.vue", 15).
ExpectLine("/*").
ExpectLine("TODO: ...").
ExpectLine("*/").
```

preslavmihaylov (changes requested):
One additional open question pending

preslavmihaylov:
> Thanks for the very helpful feedback!
> 
> I refactored the code as you pointed it out since it all seem to be very reasonable to me. Gonna push these changes in a minute.
> 
> Regarding your general question, the first jump into code was a little confusing for me. I am used to have many little unit-tests written for single methods. This is often very useful when trying to understand and especially debug code for the first time. I couldnt find any of those tests here, so I started to try and write my own little helper tests. Quickly, I noticed that the structure of the code does not "allow" this kind of tests easily since I could not even create a valid call to a method "from scratch".
> Therefore, I just started to write code "blindly" (without instant feedback), looked at the way other file-types are handled and later debugged it with some prints and the integration-test.
> Apart from the lack of these unit-tests, I find the code itself very clear and instructive, so it was possible for me to grasp the idea behind most of the things I saw, and reprocude/adapt the concepts to the new problem.

Thank you for sharing this feedback!

The reason there are no unit tests is because I decided to experiment with this project by using only integration tests & no unit tests in order to test a hypothesis I had that well-written integration tests can save you from regressions without the big cost (in my eyes) of writing & maintaining unit tests.

I hadn't considered that there'd be people like you who use the unit tests as "hooks" you can use to learn the code more easily & test new code you write in isolation.

testing/todocheck_test.go
FelixTheodor: Oh, this is weird. I mean, you are right, but I get an error if I implement it like you proposed:

`/todocheck/testing/todocheck_test.go:791: No matching todo detected in any of the scenarios
        
        Actual:
        ERROR: Malformed todo
        scenarios/vue/main.vue:0: TODO: malformed CS/JS multline entry, missing number
        scenarios/vue/main.vue:1: */
        	> TODO should match pattern - TODO {task_id}:
        
        Remaining scenarios:
        (scenario #1)
        ERROR: Malformed todo
        scenarios/vue/main.vue:15: /*
        scenarios/vue/main.vue:16: TODO: ...
        scenarios/vue/main.vue:17: */
        	> TODO should match pattern - TODO {task_id}:
        
        (scenario #2)
        ERROR: Issue is closed
        scenarios/vue/main.vue:49:   color: blue; // TODO 2: online comment wellformed in code, BUT issue closed
        `
        
And here, the displayed line of the actual todo is zero. I'm confused. Do you have any idea why it behaves that way?

FelixTheodor:
>Thank you for sharing this feedback!
>
>The reason there are no unit tests is because I decided to experiment with this project by using only integration tests & no unit tests in order to test a hypothesis I had that well-written integration tests can save you from regressions without the big cost (in my eyes) of writing & maintaining unit tests.
>
>I hadn't considered that there'd be people like you who use the unit tests as "hooks" you can use to learn the code more easily & test new code you write in isolation.

You're welcome!

This is an interesting idea. I would guess it's a trade-off; on the one hand, people already familiar with the code can add more functionality to it more quickly, but on the other hand, people who try to figure it out what's going on from scratch have a harder time (or at least miss one powerful tool for it).

testing/todocheck_test.go
preslavmihaylov: This actually is a bug with the comment matcher implementation.

There are two issues:
 * You are not matching the multiline css comments correctly. You should match `/* some todo comment */`, but instead you are matching `some todo comment */`
 * You are not tracking the linecnt at which a multiline comment begins. It should be persisted when transitioning to the multiline comment state similar to [how it's done here](https://github.com/preslavmihaylov/todocheck/blob/master/matchers/standard/comments.go#L37) and [later used here](https://github.com/preslavmihaylov/todocheck/blob/master/matchers/standard/comments.go#L89)

preslavmihaylov: You can try reproducing this by:
 * `go build todocheck` and copying the output binary to a comfortable location
 * Adding a new file `main.vue` which contains the scenario from the test
 * Add a new `.todocheck.yaml` file with contents:
```
origin: https://github.com/todocheck-tests/test-public
issue_tracker: GITHUB
```
 * run `./todocheck` & verify the results

You can try instrumenting some debug prints to see what's going on or use a standard debugger if you have one set up.

FelixTheodor: Hey, you were right, it was indeed a bug. I think the new commits I pushed fix it. At least, now all the testing works as expected. I also added another test for an invalid HTML multiline TODO, which also works. 

But there is one other thing I noticed, and I am not sure if its a feature or a bug: The last line of every file is always ignored, also by other file matchers (checked PHP briefly). This means that multiline comments that are closed in the last line are not checked for errors (since they technically never become closed). 

If this is indeed a feature, I'd be interested in an explanation for it. Maybe it would then be helpful to throw an error if the last line is not empty, since it could lead to missing TODOs otherwise?

preslavmihaylov: > But there is one other thing I noticed, and I am not sure if its a feature or a bug: The last line of every file is always 
> ignored, also by other file matchers (checked PHP briefly). This means that multiline comments that are closed in the
> last line are not checked for errors (since they technically never become closed).

The final testcase you've added:
```go
		ExpectTodoErr(
			scenariobuilder.NewTodoErr().
				WithType(errors.TODOErrTypeMalformed).
				WithLocation("scenarios/vue/main.vue", 53).
				ExpectLine("<!--").
				ExpectLine("TODO : malformed HTML multiline entry").
				ExpectLine("-->")).
```

proves it otherwise. What exactly do you mean?

If your observation is indeed correct, then this is a bug. If you can reproduce it deterministically, could you open a bug with reproduction steps & share the details?

preslavmihaylov (commented):
I think your implementation now works well and I am happy to merge it.

Several follow-ups before we close this:
 * could you verify your observation about the bug is correct? If so, could you open a bug for it with reproduction steps?
 * Could you expand the [table in the README with supported languages](https://github.com/preslavmihaylov/todocheck#supported-programming-languages)

FelixTheodor:
I think I did all the follow-ups, see the new commit and #146

preslavmihaylov (approved):

PR #156, opened 2021-07-25
Add support for all Kotlin files

### LINKED ISSUE #155: Kotlin Support
According to https://kotlinlang.org/docs/basic-syntax.html#comments, comment syntax matches that of Rust, including nested block comments, so it shouldn't be a challenge to introduce support for Kotlin / update the README table.

bengsparks:
Closes #155.

Reuses the standard comment matcher used for Rust and Swift, which supports single-, multiline comments, as well as nested comments.

### DIFF
--- a/testing/todocheck_test.go
+++ b/testing/todocheck_test.go
@@ -647,6 +647,15 @@ func TestSwiftTodos(t *testing.T) {
 			WithType(errors.TODOErrTypeMalformed).
 			WithLocation("scenarios/swift/main.rs", 7).
 			ExpectLine("/* This is /* another malformed */ TODO 3 */")).
+		// Kotlin
+		ExpectTodoErr(scenariobuilder.NewTodoErr().
+			WithType(errors.TODOErrTypeIssueClosed).
+			WithLocation("scenarios/swift/main.kt", 2).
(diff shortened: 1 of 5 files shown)


testing/todocheck_test.go
```diff
@@ -591,7 +591,7 @@ func TestGroovyTodos(t *testing.T) {
 	}
 }
 
-// Also tests Rust TODOs as rust uses the same comment syntax
+// Also tests Rust and Kotlin TODOs as they use the same comment syntax
```
preslavmihaylov, line 594: can you: 
 * include a file in `testing/scenarios/swift/` which is a kotlin source file & include todo err scenarios for it as well?
 * Could you do the same for Rust as it seems to have been missed?

preslavmihaylov (changes requested):
Apologies for the delay in reviewing your PR, for some reason I missed the nortification about the new PR & just encountered it by accident.

testing/todocheck_test.go
bengsparks: The changes requested have now been included @preslavmihaylov

preslavmihaylov (approved):

PR #158, opened 2021-08-10
Fix for youtrack issues

sattellite:
I have a panic error in version 0.4.0 for youtrack issues.

`panic: interface conversion: interface {} is nil, not bool`

Code trying use fixed index value in answer from youtrack. But my server answered with different order:

```json
{
  "customFields": [{
    "value": {
      "$type": "EnumBundleElement"
    },
    "$type": "SingleEnumIssueCustomField"
  }, {
    "value": {
      "$type": "EnumBundleElement"
    },
    "$type": "SingleEnumIssueCustomField"
  }, {
    "value": {
      "$type": "User"
    },
    "$type": "SingleUserIssueCustomField"
  }, {
    "value": null,
    "$type": "SingleUserIssueCustomField"
  }, {
    "value": {
      "$type": "VersionBundleElement"
    },
    "$type": "SingleVersionIssueCustomField"
  }, {
    "value": null,
    "$type": "DateIssueCustomField"
  }, {
    "value": {
      "isResolved": false,
      "$type": "StateBundleElement"
    },
    "$type": "StateIssueCustomField"
  }],
  "$type": "Issue"
}
```

I fixed this error with finding loop correct array element for resolve status.


sattellite:
This PR can be hotfix for latest release

issuetracker/internal/youtrack/task.go
```diff
@@ -16,7 +17,16 @@ type Task struct {
...
 func (t *Task) GetStatus() taskstatus.TaskStatus {
-	switch t.CustomFields[TaskStateIndex][Value].(map[string]interface{})[IsResolved].(bool) {
+	isResolved := false
+	// Loop through all fields to find status field
+	for _, node := range t.CustomFields {
+		if node[Type].(string) == StateType && node[Value] != nil {
+			isResolved = node[Value].(map[string]interface{})[IsResolved].(bool)
+			break
+		}
+	}
```
preslavmihaylov, line 27: In case the target state type & value is not found, we should return an error rather than silently returning that the issue is closed.

Could you modify the return of the function, associated interface & rest of implementations to return task status & error & then handle it accordingly in the caller?

The error can be:
```
couldn't fetch issue status from youtrack. This is probably on us, please file a bug report here: [link to opening an issue]
```

preslavmihaylov (changes requested):

issuetracker/internal/youtrack/task.go
```diff
@@ -16,20 +17,24 @@ type Task struct {
...
+func (t *Task) GetStatus() (taskstatus.TaskStatus, error) {
 	isResolved := false
+	var statusError error
 	// Loop through all fields to find status field
 	for _, node := range t.CustomFields {
-		if node[Type].(string) == StateType && node[Value] != nil {
+		if node[Type].(string) == StateType {
+			if node[Value] == nil {
+				statusError = errors.New("couldn't fetch issue status from youtrack. This is probably on us, please file a bug report here: https://github.com/preslavmihaylov/todocheck/issues/new?title=failed%20get%20youtrack%20issue%20status&labels=bug")
+			}
```
preslavmihaylov, line 28: just return directly from here. It's much easier than assigning this to a variable & passing it later.

On the other places, `return taskstatus.Closed/Open, nil`

preslavmihaylov (changes requested):

sattellite:
This implementation will stop checking other TODOs after error from GetStatus. This is unexpected behavior.

preslavmihaylov:
I think that's fine as if one issue fetching fails, probably all will, specifically for youtrack

preslavmihaylov (approved):
All looks good.

Thank you so much for spotting this & fixing it!

I'll add it as a patch to the current release

issuetracker/internal/youtrack/task.go
preslavmihaylov: This comment is on an old change which has already been addressed

sattellite: I started the review yesterday and forgot to submit it. Then I noticed today that I didn't submit it in. And I submit it by mistake. Sorry for the inconvenience.

PR #160, opened 2021-08-14
Nim Support

### LINKED ISSUE #157: Nim Support
Nim is one of the newer languages to gain traction, so perhaps it would be best to first discuss whether it should even be supported. 

Its single line comment syntax is identical to that of Python. However, its multiline syntax is `#[ ... ]#`, and is capable of being nested. As such, the implementation could probably be adapted from the preexisting matcher.

Supported file extensions are `.nim`, `.nims` and `.nimble`

```nim 
# An implementation of Hello World in Nim

when isMainModule: # this is more or less a compile time if
    echo "Hello World"
#[ 
Something very interesting about Nim is its decision to support Uniform Function Call Syntax.
You can inform yourself about this here: https://nim-lang.org/docs/manual.html#procedures
]#
```

bengsparks:
Closes #157.
The implementation is based on the standard nested comment variant, but substitutes `/*` and `*/` for `#[` and `]#` respectively.


bengsparks:
This PR is now ready for review @preslavmihaylov

preslavmihaylov (approved):

Issue #162, opened 2021-08-29
Extend JSON output to include linked issue & other metadata


preslavmihaylov:
Currently, the json output includes only part of the info specified by todocheck's output for closed issues.

E.g. the output for `// TODO 2: a closed issue` is:
```
[
  {
    "type": "Issue is closed",
    "filename": "main.go",
    "line": 5,
    "message": ""
  }
]
```

in this case, the linked issue is not included in the output.

## How to solve this?
Add a new field `metadata` which accepts any arbitrary string-string key-value pairs.

The reason for the new field is to enable future extensions to the messages without breaking backwards compatibility.

In this case, the whole output should look like:
```
[
  {
    "type": "Issue is closed",
    "filename": "main.go",
    "line": 5,
    "message": ""
    "metadata": {
        "issueID": "2"    }
  }
]
```

abhishekchopra0907:
hey @preslavmihaylov I would like to handle this issue could you assign me this issue

preslavmihaylov:
@abhishekchopra0907 it seems @bengsparks is already working on this. Perhaps you can check out #165

PR #164, opened 2021-08-29
Extend JSON output to include linked issue & other metadata (#162)

bengsparks:


checker/errors/errors.go
```diff
@@ -23,20 +23,27 @@ type TODO struct {
...
 	res := &struct {
-		Type     string `json:"type"`
-		Filename string `json:"filename"`
-		Line     int    `json:"line"`
-		Message  string `json:"message"`
+		Type     string   `json:"type"`
+		Filename string   `json:"filename"`
+		Line     int      `json:"line"`
+		Message  string   `json:"message"`
+		Metadata metadata `json:"metadata"`
```
preslavmihaylov, line 40: The metadata should be a `map[string]string` instead of a typed object. The reason is that different errors might have different metadata in them.

For example, `issueID` is a field applicable to non existent issue error but not for todo malformed error

checker/errors/errors.go
```diff
@@ -23,20 +23,27 @@ type TODO struct {
 	filename string
 	lines    []string
 	linecnt  int
+	issueId  string
```
preslavmihaylov, line 26: This should be a `metadata map[string]string` field similar to the json output struct

checker/errors/errors.go
```diff
@@ -71,22 +78,24 @@ func MalformedTODOErr(filename string, lines []string, linecnt int) *TODO {
...
 		filename: filename,
 		lines:    lines,
 		linecnt:  linecnt,
+		issueId:  issueId,
 	}
 }
 
 // IssueNonExistentErr when referenced todo issue doesn't exist
-func IssueNonExistentErr(filename string, lines []string, linecnt int) *TODO {
+func IssueNonExistentErr(filename string, lines []string, linecnt int, issueId string) *TODO {
```
preslavmihaylov, line 92: `issueId` should be `issueID`

checker/errors/errors.go
```diff
@@ -71,22 +78,24 @@ func MalformedTODOErr(filename string, lines []string, linecnt int) *TODO {
 }
 
 // IssueClosedErr when referenced todo issue is closed
-func IssueClosedErr(filename string, lines []string, linecnt int) *TODO {
+func IssueClosedErr(filename string, lines []string, linecnt int, issueId string) *TODO {
```
preslavmihaylov, line 81: `issueId` -> `issueID`

preslavmihaylov (changes requested):

checker/errors/errors.go
bengsparks: Changed to `map[string]string`

checker/errors/errors.go
bengsparks: Changed according to suggestion

checker/errors/errors.go
bengsparks: Completed :+1:

checker/errors/errors.go
bengsparks: Completed :+1:

preslavmihaylov:
Could you also extend the testing framework to unmarshal the metadata field and make sure it's validated in the TODO error scenarios?

bengsparks:
The newer commits now check for the `issueID` field in `metadata` part of the JSON output, and makes sure there is no such entry for malformed TODOs. @preslavmihaylov

testing/scenariobuilder/todoerr_scenario.go
```diff
@@ -48,6 +51,21 @@ func (s *TodoErrScenario) ExpectLine(line string) *TodoErrScenario {
 	return s
 }
 
+// WithMetadata extends existing metadata with a multiple of key value pairs
+// expected within the `metadata` field of the TODOs
+func (s *TodoErrScenario) WithMetadata(metadata *map[string]string) *TodoErrScenario {
```
preslavmihaylov, line 56: No need to pass a `*map[string]string`. Maps in go always behave as if they're `*map[type]type` even if you pass it as `map[type]type`

preslavmihaylov: The name of the func should be `WithJSONMetadata` as it only impacts the scenario if it is in json format

testing/scenariobuilder/todoerr_scenario.go
```diff
@@ -48,6 +51,21 @@ func (s *TodoErrScenario) ExpectLine(line string) *TodoErrScenario {
...
+	for k, v := range *metadata {
+		s.metadata[k] = v
+	}
+	return s
+}
+
+// WithMetadataEntry stores a single key-value pair that is expected in the `metadata` field
+func (s *TodoErrScenario) WithMetadataEntry(key string, value string) *TodoErrScenario {
```
preslavmihaylov, line 64: Same here in terms of naming

testing/scenariobuilder/todoerr_scenario.go
```diff
@@ -62,10 +80,29 @@ func (s *TodoErrScenario) String() string {
...
+		s.Message == other.Message &&
+		len(s.Metadata) == len(other.Metadata)
+	if !basic {
+		return false
+	}
+	for k, v := range s.Metadata {
+		if other.Metadata[k] != v {
+			return false
+		}
+	}
```
preslavmihaylov, line 103: You can extract the metadata traversal in a separate func e.g. `metadatasMatch(this, other)` and then append it to the stream of checks at line 91. That way, you can directly return the stream of checks rather than assigning it to a variable & doing this processing afterwards.

testing/scenariobuilder/validations.go
```diff
@@ -61,7 +61,7 @@ func validateJSONTodoErrs(programOutput string, scenarios []*TodoErrScenario) va
 		for _, elem := range elements {
 			removeIdx := -1
 			for j, expected := range scenarioJSONObjs {
-				if elem == *expected {
+				if elem.isEqual(expected) {
```
preslavmihaylov, line 64: a more idiomatic way for `isEqual` is `equals`

preslavmihaylov (changes requested):

testing/scenariobuilder/todoerr_scenario.go
bengsparks: `WithMetadata` -> `WithJSONMetadata`
 `*map[string]string` -> `map[string]string

testing/scenariobuilder/todoerr_scenario.go
bengsparks: `WithMetadataEntry` -> `WithJSONMetadataEntry`

testing/scenariobuilder/todoerr_scenario.go
bengsparks: Extracted metadata checks to a `func metadatasMatch(this, other *TodoErrForJSON) bool`

testing/scenariobuilder/validations.go
bengsparks: `isEqual` -> `equals`

preslavmihaylov (approved):

PR #166, opened 2021-09-04
CI: Add linter job to CI workflow

### LINKED ISSUE #165: Add a go linter CI step
The CI should run a go linter step to avoid fixing linter issues in pull requests.

Once its enabled, there would inevitably be linting issues throughout the codebase. Fix them as part of the pull request for adding the linter in CI.

TheodosiouTh:
Closes #165

### DIFF
--- a/main.go
+++ b/main.go
@@ -79,7 +79,7 @@ func main() {
 	if len(todoErrs) > 0 {
 		err = printTodoErrs(todoErrs, *format)
 		if err != nil {
-			logger.Info("Error when printing todo errors" + err.Error())
+			panic(err);
 		}
 		os.Exit(2)
 	}
--- a/main.go
+++ b/main.go
@@ -79,7 +79,7 @@ func main() {
 	if len(todoErrs) > 0 {
 		err = printTodoErrs(todoErrs, *format)
 		if err != nil {
-			panic(err);
+			panic(err)
 		}
 		os.Exit(2)
 	}
@@ -94,12 +94,12 @@ func printTodoErrs(errs []*todocheckerrors.TODO, format string) error {
 		return nil
 	}
 
-	switch format{
+	switch format {
 	case "standard":
 		for _, err := range errs {
 			fmt.Fprintln(color.Error, err.Error())
 		}
-		return nil;
+		return nil
 	case "json":
 		out := fmt.Sprintf("[%s", must(errs[0].ToJSON()))
 		for _, err := range errs[1:] {
@@ -108,7 +108,7 @@ func printTodoErrs(errs []*todocheckerrors.TODO, format string) error {
 
 		out += "]"
 		fmt.Println(out)
-		return nil;
+		return nil
 	}
 
 	return errors.New("unrecognized output format: " + format)
--- a/main.go
+++ b/main.go
@@ -79,7 +79,7 @@ func main() {
 	if len(todoErrs) > 0 {
 		err = printTodoErrs(todoErrs, *format)
 		if err != nil {
-			panic(err)
+			panic(err);
 		}
 		os.Exit(2)
 	}
@@ -94,11 +94,13 @@ func printTodoErrs(errs []*todocheckerrors.TODO, format string) error {
 		return nil
 	}
 
-	if format == "standard" {
+	switch format{
+	case "standard":
 		for _, err := range errs {
 			fmt.Fprintln(color.Error, err.Error())
 		}
-	} else if format == "json" {
+		return nil;
+	case "json":
 		out := fmt.Sprintf("[%s", must(errs[0].ToJSON()))
 		for _, err := range errs[1:] {
 			out += fmt.Sprintf(",%s", must(err.ToJSON()))
@@ -106,6 +108,7 @@ func printTodoErrs(errs []*todocheckerrors.TODO, format string) error {
 
 		out += "]"
 		fmt.Println(out)
+		return nil;
 	}
 
 	return errors.New("unrecognized output format: " + format)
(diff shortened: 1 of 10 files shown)


preslavmihaylov:
@TheodosiouTh it seems there are some more linting issues once I ran it locally.

Note that the github action you've just added doesn't run on this PR's CI so we have to make sure the check passes locally.

Pasting the output:
image

TheodosiouTh:
@preslavmihaylov Thank you for your input.
The pr is still a "work in progress" as I could not get it done yesterday.

preslavmihaylov:
Also, could you tweak the linter config a bit and include the `golint` and `gofmt` linters which are disabled by default.

Additionally, once you make that change, you can remove the gofmt step from the CI pipeline as its a subset of the step you're adding.

preslavmihaylov:
Actually, we should use `revive` as it seems `golint` is deprecated

TheodosiouTh:
@preslavmihaylov  I added `revive` and  `gofmt`  as you requested. I also removed the `gofmt` step and fixed all the linting errors and warnings. 
[Here is the final run of the workflow](https://github.com/TheodosiouTh/todocheck/runs/3517172940). 
The tests fail but I don't know why (on this PR they pass). I will keep looking at it to see what the issue is.

TheodosiouTh:
@preslavmihaylov  I have checked the code again and I do not get why the `make test` would fail after all the tests pass. My guess is that the error occurs because of the environment variables. So I think this will not be an issue when we merge the PR.

TheodosiouTh:
@preslavmihaylov  I think the PR is ready for review, feel free to check it and give me feedback.

main.go
```diff
@@ -77,7 +77,10 @@ func main() {
 	}
 
 	if len(todoErrs) > 0 {
-		printTodoErrs(todoErrs, *format)
+		err = printTodoErrs(todoErrs, *format)
+		if err != nil {
+			logger.Info("Error when printing todo errors" + err.Error())
```
preslavmihaylov, line 82: This should be a panic as we should never get here

testing/scenariobuilder/builder.go
```diff
@@ -335,7 +336,10 @@ func (s *TodocheckScenario) setupMockIssueTrackerServer() (teardownFunc, error)
 
 		for issue := range s.issues {
 			if r.URL.Path == issuetracker.IssueURLFrom(s.issueTracker, issue) {
-				w.Write(issuetracker.BuildResponseFor(s.issueTracker, issue, s.issues[issue]))
+				_, err := w.Write(issuetracker.BuildResponseFor(s.issueTracker, issue, s.issues[issue]))
+				if err != nil {
+					logger.Info("Error " + err.Error() + " occured, when writing issuetracker response of issue:" + issue)
```
preslavmihaylov, line 341: This should be a panic as it should never happen

traverser/todoerrs/todoerrs.go
```diff
@@ -32,7 +33,10 @@ func commentsCallback(chk *checker.Checker, customTodos []string, todoErrCallbac
 		if err != nil {
 			return fmt.Errorf("couldn't check todo line: %w", err)
 		} else if todoErr != nil {
-			todoErrCallback(todoErr)
+			err = todoErrCallback(todoErr)
+			if err != nil {
+				logger.Info("couldn't run todo error callback: " + err.Error())
```
preslavmihaylov, line 38: We should instead return `fmt.Errorf("received error from todo err callback: %w", err)`

traverser/comments/comments.go
```diff
@@ -42,14 +43,19 @@ func (t *Traverser) TraversePath(path string) error {
...
+				logger.Info("Error Occured during handleStateChange of file:" + filename + " at line: " + line)
+			}
 			prev = curr
 		}
 
 		curr = next
 		next = 0
-		t.handleStateChange(filename, line, linecnt, prev, curr, next)
+		err := t.handleStateChange(filename, line, linecnt, prev, curr, next)
+		if err != nil {
```
preslavmihaylov, line 56: These errors shouldn't be handled here as they are handled via the last `return t.callbackErr`.

Does the linter allow ignoring the error via `_ = t.handleStateChange(...)`?

Alternatively, you can change the signature of the function to not return an `error` and update it accordingly

preslavmihaylov (changes requested):

main.go
TheodosiouTh: I changed the code to panic if this point is reached during execution.

testing/scenariobuilder/builder.go
TheodosiouTh: I changed the code to panic if this point is reached during execution.

traverser/todoerrs/todoerrs.go
TheodosiouTh: I updated the code according to your suggestion.

traverser/comments/comments.go
TheodosiouTh: The linter does allow `_ = t.handleStageChange(...)` so I updated the code.

TheodosiouTh:
@preslavmihaylov I updated the code according to your suggestions. Also, I found out a logic bug in `printTodoErrs`, it was always returning an error even if an accepted format was provided (caused missing early return statements). I added the early return statements and changed the `if- else if` to a `switch-case` as I find switch case easier to read. You can see the bug fix on this (commit)[https://github.com/preslavmihaylov/todocheck/pull/166/commits/f0aac5aa40cb6262828b55cea10666d4e23dba88].

Also feel free to check the rest of the changes.

preslavmihaylov (approved):
Thank you for your contribution!

PR #167, opened 2021-09-06
Add support for case insensitive TODOs

Emoto13:


Emoto13:
@preslavmihaylov Ready for review

matchers/caseInsensitive/todomatcher.go
```diff
@@ -0,0 +1,39 @@
+package caseInsensitive
```
preslavmihaylov, line 1: name should be `case_insensitive` as that's the convention for package names in go.

matchers/caseInsensitive/todomatcher.go
```diff
@@ -0,0 +1,39 @@
...
+	return m.matcher.IsValid(todoToUpper(expr))
+}
+
+func (m *TodoMatcherCaseInsensitive) ExtractIssueRef(expr string) (string, error) {
+	res, err := m.matcher.ExtractIssueRef(todoToUpper(expr))
+	if err != nil {
+		return "", err
+	}
+
+	return res, nil
```
preslavmihaylov, line 33: Just `return m.matcher.ExtractIssueRef(...)` directly instead of if-else-ing

matchers/caseInsensitive/todomatcher.go
```diff
@@ -0,0 +1,39 @@
...
+	"regexp"
+
+	"github.com/preslavmihaylov/todocheck/matchers"
+)
+
+type TodoMatcherCaseInsensitive struct {
+	matcher matchers.TodoMatcher
+}
+
+func NewTodoMatcherCaseInsensitive(matcher matchers.TodoMatcher) *TodoMatcherCaseInsensitive {
```
preslavmihaylov, line 13: Should be `NewTodoMatcher` as the usage would be `case_insensitive.NewTodoMatcher`. As it is right now, usage would be very verbose

matchers/caseInsensitive/todomatcher.go
```diff
@@ -0,0 +1,39 @@
...
+	return &TodoMatcherCaseInsensitive{
+		matcher: matcher,
+	}
+}
+
+func (m *TodoMatcherCaseInsensitive) IsMatch(expr string) bool {
+	return m.matcher.IsMatch(todoToUpper(expr))
+}
+
+func (m *TodoMatcherCaseInsensitive) IsValid(expr string) bool {
```
preslavmihaylov, line 23: The name of the struct should be `TodoMatcher` so that the usage is `case_insensitive.TodoMatcher` which looks more sleek

preslavmihaylov:
Overall, very good approach. I like that you've used something similar to a decorator to achieve this as this makes it quite flexible when adding new todo matchers.

After resolving some of the minor issues I mentioned, we're good to go to merge this.

preslavmihaylov (changes requested):

PR #181, opened 2022-05-12
Implement error on unrecognized flags

### LINKED ISSUE #178: Discard unrecognised flags on program startup
if you run `todocheck help` instead of `todocheck --help` instead of getting an error, you'll get the usual app run.

In this case, some error should show up instead.

crdpa:
Now when passing any unrecognized argument, it shows an error and exits instead of running the app.

Fixes #178


preslavmihaylov (approved):
Awesome!

Thanks for the contribution

PR #184, opened 2022-07-29
Test Token instructions output

### LINKED ISSUE #130: Add tests for validating token acquisition messages
Extend the testing framework to support a new routine `ExpectOutputText` which checks that the program output contains the specified text.

Use that to write new tests which verify that the token acquisition messages for all supported issuetrackers are shown.

Example token acquisition message for github:
```
Please go to https://github.com/settings/tokens to create a read-only access token & paste it here.
Token: 
```

eisandbar:
closes #130

Added field expectedOutputText to TodocheckScenario and a method ExpectOutputText that sets it.
Calling run now checks stdout if expectedOutputText isn't empty and returns an error if stdout isn't what we expect.
Added tests for all issue trackers that  the TokenAcquisitionInstructions are what we expect.

### DIFF
--- a/testing/scenariobuilder/builder.go
+++ b/testing/scenariobuilder/builder.go
@@ -38,6 +38,7 @@ type TodocheckScenario struct {
 	deleteTokensCacheAfter bool
 	expectedExitCode       int
 	expectJSONFormat       bool
+	expectedOutputText     string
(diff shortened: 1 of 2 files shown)


preslavmihaylov (approved):
Thanks for your contribution!

Issue #83, opened 2020-10-24
Add more installation options for todocheck


preslavmihaylov:
Currently, the only way to install todocheck is via compilation from source or downloading the release binaries.

By adding additional installation options, we will increase the tool's adoption.

Create installation options for:
 * brew for Mac
 * Windows installer
 * apt-get & snap installation for Linux

PR #192, opened 2022-08-29
Add Dockerfile and docker build action

Jasper-Ben:
Add Dockerfile. Also, build, tag and push docker image on push

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -1,0 +1,51 @@
+name: docker-build
+
+on:
+  push:
+    branches:
+      - "master*"
+  release:
+    types: [published]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  docker-build:
+    name: Build and deploy container images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Export git tag
+        uses: tenhaus/get-release-or-tag@v2
+        id: tag
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            # branch event
+            type=ref,event=branch
+            # tag event
+            type=ref,event=tag
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to ghcr.io
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build image
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: true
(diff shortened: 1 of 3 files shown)


Jasper-Ben:
Delivering todocheck as container image makes the integration into CI systems easier

Jasper-Ben:
Also relates to https://github.com/preslavmihaylov/todocheck/issues/83

preslavmihaylov:
@Jasper-Ben I'm not sure this is going to work.

When I copied your `Dockerfile` locally, I did:
```
docker build . -t todocheck:test
```

Afterwards, I created a folder with `main.go`:
```
package main

import "fmt"

// TODO: invalid
func main() {
	fmt.Println("hello world")
}
```

And a `.todocheck.yaml` config:
```
origin: github.com/user/repository
issue_tracker: GITHUB
```

After I run:
```
docker run -it todocheck:test
```

I get the following error:
```
2022/09/11 07:01:10 couldn't open configuration file: file ./.todocheck.yaml not found: unable to automatically detect issue tracker: open ./.git/config: no such file or directory
```

I believe this is because todocheck is running in the context of its container, rather than my local file system.
Could you explain how this is supposed to work in CI or when used locally?

Jasper-Ben:
Hi!
Yes, you guessed the reason correctly. I should include more detail in the README file. The most straight forward way for mounting the project's directory into the container. Additionally, you would have to parse the tracker auth token into the container. When running it locally, this could look like this:

```bash
docker run -it -v /path/to/project/on/host:/project -e TODOCHECK_AUTH_TOKEN=<token>  todocheck:test --basepath /project
```

I will add this to the README.

Jasper-Ben:
Also another thing I noticed is that you add some versioning during the build of todocheck: https://github.com/preslavmihaylov/todocheck/blob/master/Makefile#L5

This currently is not the case with the docker build. I don't think it is a huge deal, since the docker tag itself can be used to identify the version, but if you say that needs to be amended, I could take a look into it.

preslavmihaylov:
I think a more sensible approach would be to configure the docker image build to be done on new releases & it should match the version of the release.

Jasper-Ben:
> I think a more sensible approach would be to configure the docker image build to be done on new releases & it should match the version of the release.

This (and more) is basically what the GitHub action will do. If you tag and push a git commit, the GitHub action will automatically build and tag the docker image appropriately. Optionally, but from experience recommended (minimum: at least for the master branch), It also builds and pushes an appropriately tagged docker image for a push to any branch, keeping the latest (head of branch). This is useful if a certain feature is required but hasn't been merged or released yet. The user of the software therefore could opt to use the image for the latest master branch commit until the next release is available.

preslavmihaylov:
I'll make a todocheck fork & play around with this to validate that it works correctly

preslavmihaylov:
Hey @Jasper-Ben I managed to play around with this in a fork & set it up nicely.

Check out https://github.com/preslavmihaylov/todocheck/pull/200

Via those changes:
 * docker image is built on push to master only + new releases
 * Running `todocheck --version` now corresponds to the git sha (if built via master) or the version (if via release)

Would you like to replicate those changes in this PR so that I can merge it? I could just merge that one, but I don't want to steal the credits :)

Jasper-Ben:
> Would you like to replicate those changes in this PR so that I can merge it?

Done :)

I've done some further cleanup to the Dockerfile (removed unnecessary `cp` and now unused make package), squashed and pushed. Please verify :)

preslavmihaylov (approved):
Thanks for the contribution!

PR #194, opened 2022-09-06
Add support for yaml files

Jasper-Ben:


preslavmihaylov:
Thanks for that!

Could you also add an integration test?

Jasper-Ben:
> Thanks for that!
> 
> Could you also add an integration test?

Done. I've added it to the "scripts" testing, as the todo comment rules are identical. If you prefer it to be differently structured no problem as well.

preslavmihaylov (approved):

PR #215, opened 2025-07-27
feat(overdue): give error on overdue issues

amirsalarsafaei:


amirsalarsafaei:
I implemented a feature which gives error when gitlab issues are in overdue state. this isn't reflected on state but can be checked by the issue due date. I implemented it in gitlab because we use gitlab but I believe it can be extended to other task managers as well. could you please review this? @presmihaylov  I have already tested this in our ci

presmihaylov (approved):

